PR #6, opened 2026-05-14
fix(agent): skip nvm bootstrap when claude/codex is already on PATH

zpzjzj:
## Summary

- `environment.type: none` runs no longer force an nvm bootstrap. The Node/nvm bootstrap is now wrapped in `if ! command -v <cli>`, so when `claude` / `codex` is already on PATH (the common host-development case) nvm is never sourced or switched.
- The bootstrap's trailing `nvm use 22` is loosened to fire only when `node_major < 22`. A system Node >= 22 is left alone instead of being forced onto v22.
- Run paths and all MCP-install builders go through the new guarded helper. Install commands themselves (`defaultClaudeCodeInstallCmd`, `defaultCodexInstallCmd`) keep the unguarded bootstrap because they own the case where the CLI really is missing.

### Why

Before this change, `skill-up run` on a host with Node v24 (no v22 installed) hard-failed with `version "v22" is not yet installed` — even though `claude` / `codex` were perfectly callable locally. That contradicts the implicit contract of `environment.type: none`, which is "use the host as-is."

## Test plan

- [x] `go test -race ./internal/agent/...`
- [x] `golangci-lint run ./internal/agent/...` — 0 issues
- [x] End-to-end via `bin/skill-up run` against the local `aone-bazel` skill with `engine: claude_code` (no model, no base_url) — PASS, transcript captured, no nvm activity in the script.
- [x] Same case with `--engine codex` — PASS.
- [x] Reverified after uninstalling Node v22 locally (only v24 present) — both engines still PASS.
- [x] New unit tests:
  - `TestNodeRuntimeCommandGuardSkipsBootstrapWhenCLIPresent` — guarded helper wraps the bootstrap in `if ! command -v 'claude' >/dev/null 2>&1; then ... fi`.
  - `TestNodeRuntimeCommandWithoutGuardKeepsBootstrap` — unguarded helper still emits the bootstrap unconditionally.

### DIFF
--- a/internal/agent/claude_code.go
+++ b/internal/agent/claude_code.go
@@ -69,7 +69,7 @@ func buildClaudeMCPInstallCmd(server runtime.MCPServerConfig) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return nodeRuntimeCommand(cmd), nil
+	return nodeRuntimeCommandWithGuard("claude", cmd), nil
 }
 
 func defaultClaudeCodeInstallCmd() string {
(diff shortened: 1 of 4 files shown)


roark47 (approved):
LGTM

PR #10, opened 2026-05-15
feat(config): support engine type in eval config

jwx0925:
## Summary
- add a release-binary install script and update installation docs
- simplify Quick Start examples and remove README project structure sections
- add engine.type support while keeping engine.name backward compatible
- add README context on how skill-up operationalizes the official Agent Skills evaluation workflow

## Tests
- bash -n install.sh
- env -u ANTHROPIC_BASE_URL make test
- env -u ANTHROPIC_BASE_URL make verify

### DIFF
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -91,6 +91,8 @@ expect:
   must_contain:
     - "Hello"
     - "World"
+  must_not_contain:
+    - "error"
 ```
 
 The case `id` defaults to the filename (`hello-world`). Add a `judge` block only when you need script-based or agent-based grading.
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -76,7 +76,9 @@ cases:
 
 > **Tip:** When `evals/eval.yaml` lives under a directory that contains `SKILL.md`, skill-up installs the current Skill automatically. The omitted fields use defaults: JSON report output, `timeout_seconds: 300`, `max_turns: 10`, and `parallelism: 1`. Add `engine.model`, `skills`, `cases.defaults`, or `report` only when you need to override them.
 
-### Step 2 — Write a case
+For the full `eval.yaml` schema, see [Writing Evals](./writing-evals).
+
+### Step 2 — Write an Eval Case
 
 Create `evals/cases/hello-world.yaml`:
 
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -6,25 +6,21 @@ skill-up is an evaluation tool for Agent Skill developers. Use it to verify that
 
 ## Installation
 
-skill-up ships as a prebuilt binary and has no runtime dependencies.
-
-### From source (recommended)
+### Install with the script
 
 ```bash
-go install github.com/alibaba/skill-up/cmd/skill-up@latest
+curl -fsSL https://raw.githubusercontent.com/alibaba/skill-up/main/install.sh | bash
 ```
 
-Or build locally from a checkout:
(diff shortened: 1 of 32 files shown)


zpzjzj:
The eval.yaml under examples/ and e2e/ may also need rename from name to type

docs/guide/getting-started.md
```diff
@@ -70,41 +66,23 @@ schema_version: v1alpha1
 environment:
   type: none                    # Plain-text Skills don't need an isolated container
 
-skills:
-  - source: local_path
```
roark47, line 74: These valid configurations should not be deleted

docs/guide/getting-started.md
```diff
@@ -70,41 +66,23 @@ schema_version: v1alpha1
...
-    path: ./dependency_skill_dir   # Relative to the Skill under test, when depending on another SKILL.md
-
 engine:
-  name: claude_code             # Use Claude Code as the Agent Engine
-  # `model` is optional. If omitted, the engine uses its local default model
-  # and `--model` is NOT passed on the command line. To pin a model, uncomment:
-  # model:
-  #   provider: anthropic
-  #   name: claude-sonnet-4-6
+  type: claude_code             # Use Claude Code as the Agent Engine
```
roark47: Use "name" instead of "type" here

docs/guide/getting-started.md
```diff
@@ -113,22 +91,16 @@ expect:
   must_contain:
     - "Hello"
     - "World"
-  must_not_contain:
```
roark47, line 116: These configuration details should not be deleted

docs/guide/writing-evals.md
```diff
@@ -45,7 +45,7 @@ environment:
   type: none
 
 engine:
-  name: claude_code
```
roark47, line 48: These changes to the name are invalid

internal/cli/run.go
```diff
@@ -400,6 +400,9 @@ func resolveEvalConfig(evalCfg *config.EvalConfig, engineName string, cmd *cobra
 	if engineName == "" {
 		if evalCfg != nil {
 			engineName = evalCfg.Engine.Name
+			if engineName == "" {
```
roark47: Here, the type field is being treated as the name field. This design addition is not needed.

roark47 (changes requested):
The changes to the installation method and the streamlined documentation are excellent. However, too many new design elements were added, which does not align with expectations. Currently, the Name field better meets the design requirements.

docs/guide/getting-started.md
jwx0925: I think `type` is better.
so I refactored this `name` to `type`

docs/guide/writing-evals.md
jwx0925: I think type is better.
so I refactored this name to type

docs/guide/getting-started.md
jwx0925: Restored the `must_not_contain` example so the quickstart still shows a negative assertion. I kept the heavier judge details out of the quickstart and left those for the full schema guide.

docs/guide/getting-started.md
jwx0925: The quickstart is intentionally kept concise so users see the smallest runnable eval config first. The removed options are still valid, but they are better covered in the full schema guide linked from this section.

internal/cli/run.go
jwx0925: This is compatibility logic. The public config now accepts `engine.type`, while the existing runtime path still reads `Engine.Name`, so we normalize the two fields before execution to keep old and new configs working consistently.

roark47 (approved):
LGTM

PR #12, opened 2026-05-15
test(e2e): verify opensandbox runtime with codex via CI sidecar

zpzjzj:
## Summary

Adds executed CI coverage for the OpenSandbox runtime. Until now the
opensandbox path never ran in CI — the existing codex opensandbox test
always skipped for lack of an external sandbox service. This PR adds a CI
job that self-hosts the OpenSandbox server on the runner, so the path is
exercised on every PR.

The opensandbox job runs the **codex** agent: claude_code already has
real-model coverage in the none-runtime e2e job, whereas codex is otherwise
only exercised against a fake binary — so this is its sole real coverage and
adds one more agent overall.

## Changes

- `.github/workflows/ci.yml`: new `e2e-opensandbox` job. Installs
  `opensandbox-server` via `uv`, starts it as a background process (Docker
  runtime, host network, host `docker.sock`), health-checks it, then runs
  `TestAgent_Codex_OpenSandboxRuntime` against `http://127.0.0.1:8080`.
- `e2e/agent_test.go`: new `TestAgent_ClaudeCode_OpenSandboxRuntime`
  (mirrors the codex test; runnable locally / in future CI) and an
  `openSandboxE2EImage()` helper that reads `OPENSANDBOX_IMAGE`.
- Both opensandbox tests preserve the in-sandbox agent workspace as a CI
  artifact for post-mortem when execution fails inside the sandbox.

## Test plan

- [ ] `make test` / `make verify` pass
- [x] `go test -tags e2e -run OpenSandbox ./e2e` passes locally (skips
      cleanly without `OPENSANDBOX_API_KEY`)
- [ ] New `e2e-opensandbox` CI job is green on this PR

## Notes for reviewers

- The sandbox image is `node:22`: skill-up bootstraps the agent CLI inside
  the sandbox (`nvm`/`node`/`npm install`), which needs `curl`/`git`/`node`.
  Bare `ubuntu:latest` lacks these. Override via `OPENSANDBOX_IMAGE`.
- The OpenSandbox server is self-hosted, so no `OPENSANDBOX_*` secret is
  needed — but the agent still calls a real model, so the job gates on
  `DASHSCOPE_API_KEY` and skips on fork PRs (same pattern as the e2e job).
- `execd_image` is pinned to `opensandbox/execd:v1.0.16`; bump if the
  server log reports an execd compatibility error.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,9 +256,9 @@ jobs:
         env:
           DASHSCOPE_API_KEY: ${{ secrets.DASHSCOPE_API_KEY }}
           DASHSCOPE_MODEL: qwen3.6-plus
-          # codex (dashscope provider) speaks the OpenAI wire API → route it
-          # through DashScope's OpenAI-compatible endpoint.
-          DASHSCOPE_BASE_URL: https://dashscope.aliyuncs.com/compatible-mode/v1
+          # codex speaks the OpenAI wire API → point it at DashScope's
+          # OpenAI-compatible endpoint via OPENAI_BASE_URL.
+          OPENAI_BASE_URL: https://dashscope.aliyuncs.com/compatible-mode/v1
           # Copies the in-sandbox agent workspace out before t.TempDir cleanup
           # so the upload step below can surface it for post-mortem.
           SKILL_UP_E2E_ARTIFACT_DIR: ${{ github.workspace }}/e2e-opensandbox-artifacts
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,17 +245,19 @@ jobs:
 
       - name: Run e2e tests (opensandbox runtime)
         if: steps.secrets.outputs.available == 'true'
-        # Only the claude_code opensandbox path is wired up here; the codex
-        # variant can be added once its provider routing is verified.
-        run: go test -tags e2e -timeout 1800s -count=1 -v -run TestAgent_ClaudeCode_OpenSandboxRuntime ./e2e
+        # Exercise the codex agent here: claude_code already has real-model
+        # coverage in the none-runtime job, whereas codex is otherwise only
+        # tested against a fake binary. This run is its sole real coverage.
+        run: go test -tags e2e -timeout 1800s -count=1 -v -run TestAgent_Codex_OpenSandboxRuntime ./e2e
         env:
-          SKILL_UP_FULL_E2E: "1"
           DASHSCOPE_API_KEY: ${{ secrets.DASHSCOPE_API_KEY }}
           DASHSCOPE_MODEL: qwen3.6-plus
-          # claude_code + dashscope provider → routed through the Anthropic-compat endpoint.
-          ANTHROPIC_API_KEY: ${{ secrets.DASHSCOPE_API_KEY }}
-          ANTHROPIC_BASE_URL: https://dashscope.aliyuncs.com/apps/anthropic
-          ANTHROPIC_MODEL: qwen3.6-plus
+          # codex (dashscope provider) speaks the OpenAI wire API → route it
+          # through DashScope's OpenAI-compatible endpoint.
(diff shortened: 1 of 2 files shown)


e2e/agent_test.go
```diff
@@ -760,6 +772,124 @@ func isOpenSandboxReadyFailure(stderr string) bool {
...
+// against a real OpenSandbox runtime. Mirrors TestAgent_Codex_OpenSandboxRuntime
+// so both supported engines have end-to-end coverage of the opensandbox bridge.
+func TestAgent_ClaudeCode_OpenSandboxRuntime(t *testing.T) {
+	// No skipIfClaudeUnavailable here: the opensandbox runtime bootstraps the
+	// claude CLI inside the sandbox, so the host runner does not need it.
+	sandboxAPIKey := openSandboxE2EAPIKey()
+	if sandboxAPIKey == "" {
+		t.Skip("OPENSANDBOX_API_KEY not set, skipping claude_code opensandbox test")
+	}
+	anthropicAPIKey := os.Getenv("ANTHROPIC_API_KEY")
```
jwx0925: The fallback to `DASHSCOPE_API_KEY` is incomplete unless the test also gets an Anthropic-compatible base URL. In the new opensandbox CI job, `DASHSCOPE_BASE_URL` is set to the OpenAI-compatible endpoint for Codex, and `Run` inherits that env. If this Claude opensandbox test is enabled there, `provider: dashscope` can resolve that OpenAI-compatible URL and pass it to `claude_code` as `ANTHROPIC_BASE_URL`, so the failure would be an endpoint mismatch rather than an OpenSandbox regression. Please either wire this test with `ANTHROPIC_BASE_URL` / the DashScope Anthropic endpoint, or isolate Codex's OpenAI-compatible `DASHSCOPE_BASE_URL` from the Claude test.

.github/workflows/ci.yml
```diff
@@ -150,6 +150,133 @@ jobs:
...
+          cache: true
+
+      - name: Install uv
+        if: steps.secrets.outputs.available == 'true'
+        uses: astral-sh/setup-uv@v8.1.0
+
+      - name: Install OpenSandbox server
+        if: steps.secrets.outputs.available == 'true'
+        run: |
+          uv tool install opensandbox-server
```
jwx0925: This installs the latest `opensandbox-server` on every CI run, while `execd` is pinned below. Since the PR notes server/execd compatibility can matter, this can make unrelated future PRs fail or silently test a different OpenSandbox behavior. Please pin `opensandbox-server` to the version validated here, or otherwise tie the server and `execd` versions together.

jwx0925 (commented):
I left two inline comments on issues that are worth addressing before merging.

e2e/agent_test.go
zpzjzj: Fixed in 8810d78: the test now pins DASHSCOPE_BASE_URL and ANTHROPIC_BASE_URL to the Anthropic-compatible endpoint (via dashScopeAnthropicE2EBaseURL) in its own run env, so the codex job's OpenAI-compatible DASHSCOPE_BASE_URL can no longer leak in.

.github/workflows/ci.yml
zpzjzj: Fixed in 8810d78: opensandbox-server is now pinned to ==0.1.13, paired in lockstep with the pinned execd image.

jwx0925 (approved):
LGTM. I re-reviewed the latest head after the follow-up commit; the previous comments are addressed and CI is green.

e2e/agent_test.go
zpzjzj: Follow-up: subsequent commits went further than the original fix. The eval `provider` is now set to each engine's native value (codex -> `openai`, claude_code -> `anthropic`) instead of `dashscope`, so skill-up resolves credentials/endpoint straight from OPENAI_* / ANTHROPIC_*. The two engines no longer share any env var, so the OpenAI-vs-Anthropic endpoint-leak surface is removed entirely rather than just isolated.

PR #29, opened 2026-05-19
fix(credential): load default credentials file in run path

zpzjzj:
## Summary

- `skill-up run` constructed the credential resolver with a hard-coded empty `confPath` (`internal/cli/run.go:216`), so `Resolver.Load()` never read a credentials file — it only honored `--api-key`, env vars, and a `.env` in the CWD.
- This contradicted the documented `~/.skill-up/credentials.yaml` fallback in `docs/guide/writing-evals.md`, `CONTRIBUTING.md`, and `AGENTS.md`.
- Added `credential.DefaultConfPath()` returning `~/.skill-up/credentials.yaml` and passed it to `NewResolver` in the run command.

## Why

The docs advertise a persistent credentials file as a fallback, but the code path made it unreachable. Users who put keys in `~/.skill-up/credentials.yaml` saw `skill-up run` fail with missing credentials.

## Notes for reviewers

- A missing credentials file is still handled gracefully — `Resolver.Load()` logs a debug line and falls through to env vars / CLI flags — so behavior is unchanged when the file is absent.
- `DefaultConfPath()` returns `""` when the home directory cannot be determined, which `Resolver` treats as "skip the file layer".
- No doc changes needed: the docs were already correct; the code was the bug.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/credential/` (added `TestDefaultConfPath`)
- [ ] Manual: place a key in `~/.skill-up/credentials.yaml` with no `--api-key`/env var and confirm `skill-up run` resolves it

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -213,7 +213,7 @@ func loadCredentialsAndAgent(cmd *cobra.Command, evalCfg *config.EvalConfig) (ag
 	cliModel, _ := cmd.Flags().GetString("model")
 	cliAPIKey, _ := cmd.Flags().GetString("api-key")
 
-	resolver := credential.NewResolver("")
+	resolver := credential.NewResolver(credential.DefaultConfPath())
 	if err := resolver.Load(); err != nil {
 		return nil, nil, credential.AgentInitParams{}, fmt.Errorf("failed to load credentials: %w", err)
 	}
--- a/internal/credential/credential.go
+++ b/internal/credential/credential.go
@@ -2,6 +2,7 @@ package credential
 
 import (
 	"os"
+	"path/filepath"
 	"sync"
 
 	"github.com/joho/godotenv"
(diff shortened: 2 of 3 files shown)


roark47:
@codex review

roark47 (approved):
LGTM

internal/cli/run.go
```diff
@@ -213,7 +213,7 @@ func loadCredentialsAndAgent(cmd *cobra.Command, evalCfg *config.EvalConfig) (ag
 	cliModel, _ := cmd.Flags().GetString("model")
 	cliAPIKey, _ := cmd.Flags().GetString("api-key")
 
-	resolver := credential.NewResolver("")
+	resolver := credential.NewResolver(credential.DefaultConfPath())
```
lijunfeng722, line 216: 不支持从命令传入吗， 直接就用的default

PR #37, opened 2026-05-20
fix(cli): make `init --config` read a source file, not a write target

zpzjzj:
## Summary

- `skill-up init --config <path>` previously meant "write the template to this path", which conflicted with the same flag's load-source meaning on `run`/`validate`/etc. Now it consistently names a **source**: `init` validates that file as a skill-up config and writes its raw bytes (comments preserved) to the target selected by `--local` (default: XDG `~/.config/skill-up/config.yaml`).
- `--local` and `--config` are no longer mutually exclusive.
- The `run`/`validate` discovery chain (`embed < user < project < --config`) is unchanged.
- Adds `userconfig.LoadFile(path)` to validate a single config file without consulting discovery layers.

### Behavior matrix

| Command | Effect |
|---|---|
| `skill-up init` | template -> `~/.config/skill-up/config.yaml` *(unchanged)* |
| `skill-up init --local` | template -> `./.skill-up.yaml` *(unchanged)* |
| `skill-up init --config foo.yaml` | reads `foo.yaml`, writes to `~/.config/skill-up/config.yaml` |
| `skill-up init --config foo.yaml --local` | reads `foo.yaml`, writes to `./.skill-up.yaml` |
| `skill-up init --config foo.yaml --print` | validates and prints `foo.yaml` to stdout |

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full suite green)
- [x] New tests in `internal/cli/init_test.go` cover: template write via `--local`, `--config` source copied verbatim (comments preserved), `--config` + `--print`, missing-source error, invalid-YAML-source error
- [x] Removed obsolete `TestInit_WriteAndForce` / `TestInit_LocalConflictsWithConfig` (asserted the old `--config = write target` semantics)
- [ ] Manual sanity: `skill-up init --config foo.yaml --local` produces a `.skill-up.yaml` that `skill-up run` then picks up via the project layer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+- `skill-up init`: `--config <path>` now names a **source** file to read
+  rather than a write target. `init` validates the source as a skill-up
+  config and writes its raw bytes (comments preserved) to the target
+  selected by `--local` (or the default XDG path). `--local` and
+  `--config` are no longer mutually exclusive. Without `--config`, `init`
+  still writes the commented template as before. This aligns `--config`
+  with the `run`/`validate` discovery chain semantics
+  (`embed < user < project < --config`).
+
+### Added
+- `userconfig.LoadFile(path)` helper for callers that need to read and
+  validate a single config file without consulting discovery layers.
+
 ## [0.1.1] - 2026-05-15
 
 ### Added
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
roark47: Version information needs to be added here

zpzjzj: Assigned 0.1.2 / 2026-05-20 (next patch from the 0.1.1 tag) in ba2b202 and added the matching release-tag link at the bottom of the file. Happy to bump differently if you'd prefer 0.2.0 given the init --config semantic change.

roark47 (approved):

PR #41, opened 2026-05-21
fix(claude_code): also write ANTHROPIC_AUTH_TOKEN for Bearer-auth proxies

zpzjzj:
## Problem

Some Anthropic-compatible proxies (typically internal `anthropic-proxy` gateways) only validate `Authorization: Bearer <token>` and reject the `x-api-key` header that the Anthropic SDK sends when only `ANTHROPIC_API_KEY` is configured.

Concretely, running `skill-up run --engine claude_code --base-url https://<internal-anthropic-proxy>/v1/anthropic-proxy --api-key <token> ...` against such a proxy fails every case with:

```
Failed to authenticate. API Error: 401 get token from header error
```

even though the token is valid — the SDK is sending `x-api-key`, the proxy only inspects `Authorization`.

## Fix

The Anthropic SDK switches to `Authorization: Bearer` when `ANTHROPIC_AUTH_TOKEN` is set, and the official Anthropic API also accepts Bearer. Writing both env vars in `ClaudeCodeAgent.Run` covers both proxy flavours without forcing operators to maintain two credentials or branch on proxy type.

Concrete changes:

- `internal/credential/credential.go` — add `EnvAnthropicAuthToken = "ANTHROPIC_AUTH_TOKEN"` constant alongside the existing `EnvAnthropicAPIKey` / `EnvAnthropicBaseURL`.
- `internal/agent/claude_code.go` — in `Run`, populate `ANTHROPIC_AUTH_TOKEN` alongside `ANTHROPIC_API_KEY` whenever `Cfg.APIKey` is set.
- `internal/agent/claude_code_test.go` — new test `TestClaudeCodeRun_WritesBothAnthropicAuthEnvVars` asserts both env vars land in the exec env.

## Why both, not "switch to AUTH_TOKEN"

Setting both is safe: `ANTHROPIC_AUTH_TOKEN` takes precedence in the SDK (sends Bearer), and the official Anthropic endpoint accepts both header forms. Switching solely to `AUTH_TOKEN` would risk regressions for any proxy / SDK consumer that only inspects `x-api-key`.

## Verified

- `go test ./internal/agent/... ./internal/credential/...` passes locally.
- End-to-end: applied the same change in our component layer (`skill-up` CI component) and observed the upstream call go from `401 get token from header error` → `200` against the proxy in question, confirming Bearer auth now works.

## Scope

Smallest possible change. No public API change. Only adds an extra env var on the claude_code Run path.

### DIFF
--- a/internal/agent/claude_code.go
+++ b/internal/agent/claude_code.go
@@ -102,6 +102,15 @@ func (a *ClaudeCodeAgent) Run(ctx context.Context, rt Runtime, opts ExecOptions,
 	logging.DebugContextf(ctx, "Session ID: %s", sessionID)
 
 	envVars := a.credentialEnvVars(credential.EnvAnthropicAPIKey, credential.EnvAnthropicBaseURL)
+	// Some Anthropic-compatible proxies (e.g. internal anthropic-proxy
+	// gateways) only validate `Authorization: Bearer <token>` and do not
+	// recognise the `x-api-key` header. The Anthropic SDK switches to
+	// Bearer when ANTHROPIC_AUTH_TOKEN is set, and the official Anthropic
+	// API also accepts Bearer, so writing both env vars covers both
+	// flavours without forcing operators to maintain two credentials.
+	if a.Cfg.APIKey != "" {
+		envVars[credential.EnvAnthropicAuthToken] = a.Cfg.APIKey
+	}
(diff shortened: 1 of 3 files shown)


lijunfeng722 (commented):
LGTM

lijunfeng722 (approved):

PR #42, opened 2026-05-21
fix(cli): gate --model provider/name split on provider being configured

zpzjzj:
> Replaces the original per-agent `effectiveModelName` reconstruction with a CLI-level split decision. See [revert of the previous approach](../commits/b4f8f57) and the [new commit](../commits/c670ff9). Bot review feedback addressed in-thread.

## Problem

The CLI parser at `internal/cli/run.go` unconditionally splits `--model provider/name` into `ModelProvider` + `ModelName`. After the split, two distinct user intents become indistinguishable:

1. **Credential-namespace usage** (works today): `provider: dashscope, name: claude-sonnet-4-6` — `dashscope` selects credentials, the bare model name is what hits the wire against dashscope's Anthropic-compatible endpoint.

2. **Literal model identifier**: `--model anthropic_modelscope/deepseek-v4-pro` against an internal anthropic-proxy gateway (e.g. ducky's `/v1/anthropic-proxy`) that registers models under `provider/name` keys. The full string IS the model id, but skill-up's eager split strips it so claude sees only `deepseek-v4-pro` → `400 model not found`.

Reproduced:

```
# direct, works
claude -p \"ping\" --model anthropic_modelscope/deepseek-v4-pro
# → pong

# via skill-up, fails
skill-up run --engine claude_code --model anthropic_modelscope/deepseek-v4-pro ...
# → API Error: 400 {\"error\":\"model not found\"}
```

## Fix

Resolve at the CLI / credential boundary, **before** the agent sees ModelProvider+ModelName:

- After the resolver is loaded in `loadCredentialsAndAgent`, ask whether the would-be provider is actually configured.
- **Configured** (entry in credentials.yaml OR `<PROVIDER>_API_KEY` / `<PROVIDER>_BASE_URL` env): keep the split. Existing dashscope/routify behaviour unchanged; claude_code, codex, qodercli all see the same shape they did before. codex still drives `runProviderConfig` off ModelProvider.
- **Unconfigured**: collapse. `ModelProvider` becomes `\"\"`, `ModelName` becomes the literal `provider/name` string. The agent forwards it verbatim. ducky and other anthropic-proxy gateways now reach claude with the full identifier intact.

No `effectiveModelName` changes in any agent. Each agent keeps its existing assumption: when ModelProvider is non-empty, it's a real configured provider; ModelName is always the literal model identifier to use.

## Implementation

- `internal/credential/agent_init.go` — new exported `ProviderConfigured(provider, resolver) bool`. Resolver lookup OR `<PROVIDER>_API_KEY|BASE_URL` env probe. Docstring spells out the two interpretations it disambiguates.

- `internal/cli/run.go`:
  - new `collapseUnconfiguredProviderSplit(evalCfg, resolver)` runs in `loadCredentialsAndAgent` right after `resolver.Load()`, before `ResolveRunnerInitParams`. Safe no-op when Provider is empty.
  - `normalizeCLIModelOverride` now takes a resolver and only peels off the provider prefix when the provider is configured. Without this, the prefix would be re-stripped right after the un-split landed.

## Tests

- `ProviderConfigured` (credential pkg): empty / resolver-known / `<PROVIDER>_API_KEY` env / `<PROVIDER>_BASE_URL` env / fully unconfigured.
- `collapseUnconfiguredProviderSplit` (cli pkg): unknown provider → collapses to `anthropic_modelscope/deepseek-v4-pro`; configured `dashscope` → keeps split as `dashscope` + `claude-sonnet-4-6`; empty-provider → no-op.
- `normalizeCLIModelOverride`: extra case `anthropic_modelscope/deepseek-v4-pro` is preserved end-to-end when the provider is unconfigured.

`go test ./...` passes locally.

## Why this approach over per-agent fixes

Earlier iteration of this PR reconstructed `provider/name` inside claude_code's `effectiveModelName` (skipping the agent-default `anthropic` provider). That looked surgical but introduced a regression for `provider: dashscope, name: claude-sonnet-4-6` — credential-namespace usage was silently rewritten to `dashscope/claude-sonnet-4-6` on the wire. The two intents are simply indistinguishable after the CLI split, so any per-agent reconstruction picks the wrong one for half the population.

Gating at the CLI / credential boundary uses the only signal that actually separates the two cases: whether the user has supplied a configuration path for the provider name. Configured = namespace; unconfigured = literal.

Engine-agnostic, so a new agent later doesn't need to re-derive the same workaround.

## Pairs with #41

PR #41 fixes the auth side of the same anthropic-proxy scenario (Bearer token routing). This PR fixes the model-routing side. Either can be reviewed and merged independently.


internal/agent/claude_code.go
```diff
@@ -152,7 +152,23 @@ func (a *ClaudeCodeAgent) Run(ctx context.Context, rt Runtime, opts ExecOptions,
...
+	// `--model provider/name` is split into ModelProvider+ModelName by the CLI
+	// parser. For the default Anthropic endpoint, the provider prefix is just
+	// routing metadata and must not reach claude (the real API only knows bare
+	// model names). For custom Anthropic-compatible proxies, however, the
+	// provider prefix is part of the model identifier the proxy registers
+	// against (e.g. ducky's `anthropic_modelscope/deepseek-v4-pro`); stripping
+	// it leaves the bare name and the proxy returns 400 model not found.
+	// Reconstruct provider/name iff the provider is non-default.
+	provider := strings.TrimSpace(a.Cfg.ModelProvider)
+	if provider == "" || provider == agentProviderAnthropic {
```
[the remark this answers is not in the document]
zpzjzj: Thanks — agreed. In the refactor I'm pushing next, this comparison goes away: the CLI no longer eagerly assumes a slashed `--model` produces a custom provider. Instead it falls back to bare-name handling whenever the would-be provider isn't actually configured (resolver entry / `<PROVIDER>_API_KEY|BASE_URL` env). For the official Anthropic path that means `ANTHROPIC_API_KEY` being set already makes the provider "configured" → split keeps working regardless of case, and claude_code receives bare ModelName as before.

internal/agent/claude_code.go
```diff
@@ -152,7 +152,23 @@ func (a *ClaudeCodeAgent) Run(ctx context.Context, rt Runtime, opts ExecOptions,
...
+		return name
+	}
+	return provider + "/" + name
```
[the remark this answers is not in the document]
zpzjzj: Good catch — the same refactor removes the unconditional concat: claude_code goes back to returning bare `Cfg.ModelName` as upstream did before this PR. The `<PROVIDER>_MODEL` env path therefore reaches the agent untouched, no double-prefix possible.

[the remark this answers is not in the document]
zpzjzj: Agreed, this is a real regression and the design needs to change. `ModelProvider` is a credential-namespace identifier and after a CLI split, `provider: dashscope, name: claude-sonnet-4-6` and `--model anthropic_modelscope/deepseek-v4-pro` look identical — but the user intent is opposite (former: provider is namespace, bare name on the wire; latter: full string IS the model id).

Pushing a redesign: gate the CLI split on "is this provider actually configured?" (resolver entry / `<PROVIDER>_API_KEY|BASE_URL` env). When yes, split as today and claude_code emits bare name (no regression for dashscope-claude). When no, treat the whole slashed string as the opaque model id and claude_code passes it through verbatim (the ducky `anthropic_modelscope/deepseek-v4-pro` case still reaches claude intact). claude_code's `effectiveModelName` reverts to the original bare `Cfg.ModelName` so the agent no longer has to guess.

Will push shortly and update the PR description.

internal/cli/run.go
```diff
@@ -401,6 +411,36 @@ func evaluateOptionsFromFlags(cmd *cobra.Command) (runner.EvaluateOptions, error
...
+// expects verbatim (e.g. an internal anthropic-proxy gateway that registers
+// models under `anthropic_modelscope/deepseek-v4-pro`) than a credential
+// namespace prefix the agent should peel off.
+//
+// Configured providers (e.g. `dashscope` with DASHSCOPE_API_KEY set) keep
+// the split — that's the existing namespace-style usage where the bare
+// model name is what should hit the wire.
+//
+// Safe no-op when Provider is empty or already collapsed.
+func collapseUnconfiguredProviderSplit(evalCfg *config.EvalConfig, resolver *credential.Resolver) {
```
zpzjzj: **🟠 Major: 「split + ProviderConfigured」判定逻辑分散三处 + 静默回收无可观测性**

同一个「provider/name 是不是 namespace 前缀」的判定被实现了三次：
- `resolveEvalConfig`（run.go:461-469）做乐观拆
- `collapseUnconfiguredProviderSplit`（本函数 428-442）反向回滚
- `normalizeCLIModelOverride`（run.go:597-610）对 cliModel 第三次独立判断

未来任何口径变化（如允许 `provider:name` 这种新分隔、增加 disable 语义）都得同步改三处，极易遗漏。**「先拆 → 再补回」本身就是反模式**。

另外当用户把 provider 拼错（如 `dashsope/...`）时，代码静默把整串当模型 id 透传，最终拿到不直观的 400，且无 debug log。同样问题在 `agent_init.go:ProviderConfigured` —— shell 里随手 export 的 `FOO_API_KEY` 会让 `--model foo/bar` 的语义在两台机器上反转。

**建议**：
1. 把决策下沉到 credential 包，暴露 `credential.ResolveModelRef(s string, r *Resolver) (provider, name string)`；CLI 只在 resolver 加载后调用**一次**，统一改写 `evalCfg.Engine.Model` 和 cliModel；`resolveEvalConfig` 不再做乐观拆分（或仅做语法校验），从根本消掉「先拆再合」反模式
2. 命中 collapse 分支时打 debug 日志：`treating "%s" as opaque model id; provider %q has no API key / base URL configured`，便于排障

internal/cli/run.go
```diff
@@ -551,13 +591,21 @@ func formatModelRef(provider, name string) string {
 	}
 }
 
-func normalizeCLIModelOverride(modelFlag string) string {
+func normalizeCLIModelOverride(modelFlag string, resolver *credential.Resolver) string {
 	if modelFlag == "" {
```
zpzjzj, line 612: **🟠 Major: `normalizeCLIModelOverride` 与 `collapseUnconfiguredProviderSplit` 是同一决策的第三处实现 + 决策与 engine 解耦的稳定性**

本函数与 `collapseUnconfiguredProviderSplit`（line 428）做完全相同的「split + ProviderConfigured」判定，只是数据源换成 cliModel。两个函数注释相互引用（"same disambiguation as"），但实现各自维护一份 SplitN + ProviderConfigured 逻辑 + 8 行 rationale 注释。

**架构层面的稳定性顾虑**：新方案把决策从 agent 调用点（每次执行）提前到 CLI 边界（一次性），优点是 `ModelName/Provider` 在 run 生命周期内不变；缺点是丢失了 per-agent 判定上下文 —— 例如 `codex.go:264-275` 的 `codexCustomProviderUnavailableReason` 对 `provider/` 前缀的可用性判断与 claude_code 不同。当前 CLI 不知道 engine，所有 engine 共用同一份判断。

**建议**：
1. 与 line 428 一起抽到 `credential.ResolveModelRef`，统一三处逻辑（含 resolveEvalConfig 那次）
2. `ResolveModelRef` 签名留出 `engine` 参数作扩展点（即便当前 ignore），或在 godoc 明确「此判定与 engine 无关」的约束
3. 8 行 rationale 注释只在 `ResolveModelRef` 的 godoc 里写一次，调用处一行 `// see ResolveModelRef`

internal/credential/agent_init.go
```diff
@@ -234,6 +234,41 @@ func lookupProviderEnv(provider string, kind scopedValueKind) (value, envVar str
...
+//     credential namespace (e.g. `provider: dashscope, name: claude-sonnet-4-6`
+//     uses dashscope's API key but talks to a bare Anthropic model id) and a
+//     split is appropriate.
+//   - When the provider is NOT configured, the slashed string is more likely
+//     a literal model identifier the upstream API expects verbatim
+//     (e.g. `anthropic_modelscope/deepseek-v4-pro` registered as-is on an
+//     internal anthropic-proxy gateway). Callers can treat it as opaque.
+//
+// Returns false for empty provider input.
+func ProviderConfigured(provider string, resolver *Resolver) bool {
```
zpzjzj: **🟠 Major: `ProviderConfigured` 接口设计问题（类型泄漏 + 职责错位 + env 静默 fallback）**

三个独立问题汇总：

1. **类型泄漏**：签名 `ProviderConfigured(provider string, resolver *Resolver) bool` 强制 caller 依赖具体 `*Resolver`，难 mock。
2. **职责错位**：函数实际语义是「provider 有没有可用配置来源」，但住在 `agent_init.go`（文件主题是构造 `AgentInitParams`），与文件职责无关。
3. **混合两种 configured 语义**：同时检 `resolver.Get()` 和 `lookupProviderEnv(env API_KEY/BASE_URL)`，调用方无法只看其中一种。用户在 shell 里为另一个工具 export 过 `FOO_API_KEY`，再跑 `--model foo/bar` 时**语义会静默反转**（按 namespace 拆 vs 当字面量传），且没有任何日志可追溯。

**建议**：
1. 改为 `(r *Resolver) HasProvider(name string) bool` 方法（或接口入参 `type providerLookup interface{ Get(string) (*config.APIKeyConfig, bool) }`），env fallback 由 Resolver 内部封装
2. 抽到独立文件 `provider_query.go` 或合并到 [[ResolveModelRef]] 设计中
3. 在 env fallback 命中时打 debug log：「peeling provider X because configured via env Y」

zpzjzj (commented):
## 代码评审报告（自动）

| 项目 | 内容 |
|------|------|
| **评审人** | @zpzjzj (via Claude Code github-code-review skill) |
| **评审分支** | `fix/claude-code-preserve-model-provider` → `main` |
| **HEAD** | `c670ff94` |
| **规模** | 4 files, +228 / -6（其中 145 行测试） |

### 问题汇总

- 🔴 **Blocker: 0**
- 🟠 **Major: 8**（3 条行评论 + 5 条 summary）
- 🔵 **Info: 6**

### 整体评价

**⚠️ 有条件合并** —— 无 Blocker，bug fix 的逻辑方向是对的（决策点提到 CLI/credential 边界比上一版 agent 内部 `effectiveModelName` 更稳定）。但当前实现把同一决策散到三处，缺少可观测性日志，且文档完全没动。

CI 评审时 5 个 check 全部 QUEUED（CI run #26218814647），唯一已完成的 `license/cla` 通过。请在 CI 跑完后回看。

### 关键改进方向

1. **消除三处重复决策**：抽 `credential.ResolveModelRef(s, resolver)`，消掉 `resolveEvalConfig` 乐观拆 → `collapseUnconfiguredProviderSplit` 回滚 → `normalizeCLIModelOverride` 第三次判断 这套「先拆再合」反模式（行评论 1+2）。
2. **可观测性补漏**：collapse 命中分支必须有 debug log，否则用户拼错 provider 时只能拿到 400。同时 `agent_init.go:187-196` 的 span/metric 快照早于 collapse，会上报错误的 model 维度（见下）。
3. **接口收敛**：`ProviderConfigured` 改为 `Resolver.HasProvider()` 方法，去掉类型泄漏 + 文件错位 + env 静默反转（行评论 3）。
4. **文档必须更**：CHANGELOG + cli-reference + `--model` flag help string 三处都漏（见下）。

### 行评论列表（已发布）

| # | 文件:行 | 严重度 | 问题 |
|---|---------|--------|------|
| 1 | `cli/run.go:428` | 🟠 | 三处重复 split + 静默回收无可观测性 |
| 2 | `cli/run.go:595` | 🟠 | normalizeCLIModelOverride 第三处重复 + per-engine 稳定性 |
| 3 | `credential/agent_init.go:254` | 🟠 | ProviderConfigured 类型泄漏 + 职责错位 + env 静默 fallback |

### Major（行评论无法挂载，PR 未触碰原行）

- **🟠 可观测性快照早于决策** — `internal/cli/run.go:187-196`：`formatModelRef + span.SetAttributes + observability.RecordRunStarted + ui.Statusf` 在 collapse 之前发出。对 `--model anthropic_modelscope/deepseek-v4-pro` 这类场景会上报 `provider=anthropic_modelscope, model=deepseek-v4-pro`，与最终下发给 agent 的字符串不一致，按 model 聚合的指标会错位。
  建议：把 187-196 这段挪到 `loadCredentialsAndAgent` 之后，或在 collapse 后再 refresh 一次 attribute / UI 提示。

- **🟠 `--model` flag help string 与新行为不一致** — `internal/cli/run.go:103`："accepts either a bare model name or provider/name"，未表达 gating；用户从 `--help` 看不到关键判定条件。
  建议改为："...provider/name (split only when provider has credentials/env configured; otherwise treated as a literal model id)"。

- **🟠 CHANGELOG.md 未更新** — `CONTRIBUTING.md:25` 明确要求 user-visible 变更须更新。`--model provider/name` 的拆分语义由"无脑拆"变为"只在 provider 配置过时才拆"，对依赖旧行为的用户是 breaking。建议在 `## [0.1.2] - 2026-05-21 / ### Fixed` 加一条，举出 `anthropic_modelscope/deepseek-v4-pro` 现在能透传作为代表场景。

- **🟠 `docs/guide/cli-reference.md` 与中文版描述失真** — line 32 当前写「Override model (format: `provider/name`)」，未反映新规则；line 50 示例 `--model openai/gpt-4` 隐含 openai 已配置的前提，未说明。建议改为「`provider/name`（当 `provider` 已配置 credential / `<PROVIDER>_API_KEY` / `<PROVIDER>_BASE_URL` 时按 namespace 拆分；否则整串作为字面 model id 透传）」。

- **🟠 与 engine 解耦的稳定性顾虑** — 新方案丢失 per-agent 判定上下文（`codex.go:264-275` 的 `codexCustomProviderUnavailableReason` 对 `provider/` 的可用性判断与 claude_code 不同）。当前 CLI 不知道 engine。建议在 `ResolveModelRef` 签名上留 engine 参数扩展点。

### Info 级建议

代码：
- `collapseUnconfiguredProviderSplit` docstring `Safe no-op when Provider is empty or already collapsed.` —— 两个条件在本函数语义下等价，并列易让读者误以为还有另一种 collapsed 状态需要检测；建议合并表述
- `run.go:434` 的 `name == ""` 防御分支永远不会命中（上游 `resolveEvalConfig:463` 只在 `parts[0] != "" && parts[1] != ""` 时才写入），可加一行注释说明「纯防御，理论上不可命中」
- 测试覆盖缺一个集成 case：`cliModel` 与 `evalCfg.Engine.Model` 同时存在且 provider 未配置时 `normalize` 与 `collapse` 协同生效

文档：
- **revert 历史未在代码注释中留痕**（中等严重）：comments 解释了「为什么 gating」，但未提及上一版 commit 632abc3（agent 侧 preserve prefix 方案）已被 b4f8f57 revert，后续维护者可能再次尝试在 agent 端解决绕开本 PR 设计。建议在 `collapseUnconfiguredProviderSplit` 头加一句 "See reverted commit 632abc3: an earlier agent-side preserve-prefix attempt was insufficient because the split happens before the resolver loads."
- `docs/guide/user-config.md` 缺 provider/credentials 概念入口，用户无处了解 "provider 作为 credential namespace" vs "literal model id" 的语义；建议补一小节 "Model reference syntax in `--model` / `engine.model`"

CI：
- 评审时 Build & Test / Lint / E2E ×2 / GoReleaser Check 全 QUEUED；待完成后回查

### 已具备的好实践

- 决策点选址正确（CLI/credential 边界 vs agent 内部，比上一版 effectiveModelName 稳定）
- `ProviderConfigured` 用「resolver + env fallback」两层兜底，覆盖纯 env 配置场景
- 三个核心 case（已配置 / 未配置 / 空 provider）都有单测覆盖
- 注释充分解释「为什么 gating」（即便位置 / 复制次数可优化）

---

**说明**：本评审由 4 个并行子 Agent（代码质量、架构、CI、文档）完成。性能 / 安全维度因本 PR 无相关攻击面/热路径，跳过。如对结论有异议，欢迎在对应行评论下回复。

internal/cli/run.go
zpzjzj: Addressed in commit 7f7a3c7. The SplitN + provider-config probe is now centralised in `credential.ResolveModelRef`; `collapseUnconfiguredProviderSplit` and `normalizeCLIModelOverride` both delegate to it, and `resolveEvalConfig`'s tentative split is what ResolveModelRef rolls back when the resolver later disagrees. Debug log on the collapse path lands when verbosity is high enough — typo'd providers now leave a breadcrumb instead of a silent 400.

The "先拆 → 再合" pattern itself stays for the moment because `resolveEvalConfig` runs before the resolver is loaded; eliminating it entirely needs a `ModelConfig.RawRef` field (defer the split until resolver-time) which is a wider refactor than this PR. Open to doing it as a follow-up.

internal/cli/run.go
zpzjzj: Addressed in commit 7f7a3c7 alongside the sibling thread: the SplitN-and-ProviderConfigured logic only lives in `credential.ResolveModelRef` now; both `collapseUnconfiguredProviderSplit` and this function delegate to it, and the rationale block is consolidated into ResolveModelRef's godoc with one-line `// See credential.ResolveModelRef` pointers at the call sites.

On the engine-awareness concern: ResolveModelRef's signature stays engine-free for now (every supported engine agrees that "stored config exists" is the right signal), but the godoc explicitly calls out the extension point in case a future engine needs a different policy.

internal/credential/agent_init.go
zpzjzj: Addressed in commit 7f7a3c7 — all three points:

1. **Type leak**: replaced standalone `credential.ProviderConfigured(provider, *Resolver)` with `(r *Resolver) HasProvider(name) bool` method. Callers pass the Resolver they already have; tests use the same shape (including a nil receiver, which is intentionally safe and exercised by `TestHasProvider_EmptyNameReturnsFalse` / `TestHasProvider_NilReceiverFallsBackToEnv`).
2. **Job placement**: moved to a new file `internal/credential/provider_query.go` so `agent_init.go` keeps its narrow focus on `AgentInitParams` construction.
3. **Silent env fallback**: ResolveModelRef now emits a debug log on the collapse path (`treating %q as opaque model id; provider %q has no resolver entry / %s_API_KEY / %s_BASE_URL / %s_PERSONAL_ACCESS_TOKEN env`). The configured-split path stays silent (it's the common case) but the disambiguation is now observable at `--verbose` time.

internal/cli/run.go
```diff
@@ -551,13 +618,21 @@ func formatModelRef(provider, name string) string {
...
-		return parts[1]
+		// Only peel off the provider prefix when it's a configured provider
+		// — same disambiguation as collapseUnconfiguredProviderSplit. For
+		// unknown providers the slashed string is more likely a literal
+		// model identifier the upstream API expects verbatim (e.g. ducky's
+		// `anthropic_modelscope/deepseek-v4-pro`); stripping would defeat
+		// the un-split applied to evalCfg.Engine.Model.
+		if credential.ProviderConfigured(parts[0], resolver) {
+			return parts[1]
+		}
```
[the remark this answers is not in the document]
zpzjzj: Right — fix in commit 26239e0. `normalizeCLIModelOverride` now takes `cliAPIKey` and `evalBaseURL` and delegates to a shared `resolveModelRefWithCLIHints` helper that mirrors the same short-circuits `collapseUnconfiguredProviderSplit` already had. The two call sites stay in lockstep so applyCLIOverrides can never receive a model identifier that contradicts the post-collapse evalCfg state. New `TestNormalizeCLIModelOverride_CLIAPIKeyKeepsSplitStripping` covers the regression directly.

internal/cli/run.go
```diff
@@ -401,6 +411,63 @@ func evaluateOptionsFromFlags(cmd *cobra.Command) (runner.EvaluateOptions, error
...
+	// separate YAML keys) reach here without `/` in cliModel, and must be
+	// preserved.
+	if !strings.Contains(cliModel, "/") {
+		return
+	}
+	if cliAPIKey != "" {
+		return
+	}
+	if credential.ProviderConfigured(provider, resolver) {
+		return
```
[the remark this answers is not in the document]
zpzjzj: Fix in commit 26239e0. `credential.Resolver.HasProvider` now returns true unconditionally for the framework defaults (`anthropic`, `openai`, case-insensitive): their upstream APIs only accept bare model ids, so a slashed `--model X/Y` with X ∈ {anthropic, openai} can only mean namespace usage, and the underlying agent CLIs (claude, codex) carry their own persisted login state that the resolver/env probe can't see. The `--model anthropic/X` + login-state-only flow stops being collapsed. Covered by `TestHasProvider_TrueForFrameworkDefaults` and `TestCollapseUnconfiguredProviderSplit_KeepsAnthropicSplitWithoutEnv`.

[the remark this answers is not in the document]
zpzjzj: Fix in commit 26239e0. Both `collapseUnconfiguredProviderSplit` and `normalizeCLIModelOverride` now go through `resolveModelRefWithCLIHints`, which honours `evalBaseURL` (sourced from `evalCfg.Engine.Model.BaseURL`) as a provider-configuration signal alongside `cliAPIKey`. When the user sets `engine.model.base_url` in eval.yaml the split is preserved regardless of resolver/env state. Covered by `TestCollapseUnconfiguredProviderSplit_KeepsSplitWhenEvalBaseURLProvided` and `TestNormalizeCLIModelOverride_EvalBaseURLKeepsSplitStripping`. Kept the signal at the CLI/eval layer rather than baking into `HasProvider` so `credential.ResolveModelRef` itself stays pure (persisted/env-resident config only).

internal/cli/run.go
```diff
@@ -401,6 +411,63 @@ func evaluateOptionsFromFlags(cmd *cobra.Command) (runner.EvaluateOptions, error
...
+//   - evalBaseURL: the user has set `engine.model.base_url` in eval.yaml,
+//     which is itself a concrete provider-configuration signal (the
+//     gateway endpoint is bound to that provider namespace).
+//
+// When either signal is present and `modelFlag` has the shape
+// `prefix/name`, the split is kept regardless of resolver/env state.
+// Otherwise falls through to ResolveModelRef.
+func resolveModelRefWithCLIHints(modelFlag string, resolver *credential.Resolver, cliAPIKey, evalBaseURL string) (provider, name string) {
+	if cliAPIKey != "" || evalBaseURL != "" {
+		parts := strings.SplitN(modelFlag, "/", modelFormatParts)
```
zpzjzj: **🟠 Major (v2): CLI hint 分支再次手写 split，仍存在「重复决策」**

第一轮意见 [[第一轮 #1+#2]] 抽 `credential.ResolveModelRef` 是对的，但这里 `resolveModelRefWithCLIHints` 在 `cliAPIKey != "" || evalBaseURL != ""` 时短路又写了一遍 `SplitN(modelFlag, "/", modelFormatParts)`，与 `provider_query.go:91-94` 是同一份逻辑。

加上 `resolveEvalConfig:488-497` 的乐观拆分、`parseJudgeModel`（`agent_init.go:103`）的判官侧拆分，**全局仍有 4 处 SplitN**，「三处重复」的根源没彻底消除，只是被 ResolveModelRef 包装了。

另外这个 hints 分支**完全旁路 HasProvider**，无任何校验或 warn。`--api-key sk-x --model totally-fake/x` 会把 `totally-fake` 当 provider 名，applyCLIOverrides 把 sk-x 塞给它，下游 provider 路由可能误诊，用户难定位。

**建议**：
1. 把 `splitProviderName(raw) (provider, name string, ok bool)` 抽到 `credential` 包，或让 `ResolveModelRef` 增加 `extraConfigured []string` 参数把 CLI 信号一起喂进去 —— 所有 split 走唯一函数
2. hints 分支命中时 Warn：`provider %q implicit from --api-key/base_url, not configured elsewhere`，给用户排查抓手

internal/credential/provider_query.go
```diff
@@ -0,0 +1,104 @@
...
+// callers that need a provider check before any resolver has been loaded.
+//
+// Used to disambiguate the two valid interpretations of a slashed
+// `--model provider/name` input — see ResolveModelRef.
+func (r *Resolver) HasProvider(name string) bool {
+	if name == "" {
+		return false
+	}
+	switch strings.ToLower(name) {
+	case "anthropic", "openai":
```
zpzjzj: **🟠 Major (v2): `HasProvider` 硬编码 framework defaults + 语义双重身份**

两个独立问题：

1. **credential 层反向依赖 engine 概念**：`HasProvider` 硬编码 `"anthropic"` / `"openai"` 总为 true（即所谓 framework-default providers）。doc 自承「Engine-awareness: not currently required」—— 但事实上这两个名字是 framework/engine 的知识，新增第三个 engine 默认时极易遗忘改这一处。
2. **函数语义双重身份**：`HasProvider` 同时回答「凭据配置存在」和「无条件视为已配置」两件事，doc 解释长达 24 行。`true` 不等于 "resolver/env 里有"，调用方读取时容易混淆。

**建议**：
1. 把 framework defaults 改成 `Resolver` 字段（`frameworkDefaults []string`），由 `NewResolver` 构造时注入；新增 engine 时无需改 `credential` 包
2. 拆成两个函数：`HasProviderCredential` + `IsFrameworkDefault`，或重命名 `ShouldKeepProviderSplit` 让职责显式
3. 配套：`collapseUnconfiguredProviderSplit` 名字也误导 —— 它的行为是「基于完整信号 reconcile，可能保留 split 也可能撤销」，注释里只描述 undo 视角；建议改 `reconcileProviderSplit` 或 `recheckProviderSplit`

internal/credential/provider_query.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+	if _, _, ok := lookupProviderEnv(name, valueAPIKey); ok {
+		return true
+	}
+	if _, _, ok := lookupProviderEnv(name, valueBaseURL); ok {
+		return true
+	}
+	if v := os.Getenv(strings.ToUpper(name) + "_PERSONAL_ACCESS_TOKEN"); v != "" {
+		return true
```
zpzjzj, line 71: **🟠 Major (v2): PAT env 名拼接绕过了归一化，破折号 provider 名会失效**

`strings.ToUpper(name) + "_PERSONAL_ACCESS_TOKEN"` 直接拼接：若 provider 名带破折号（如 `anthropic-proxy`），生成的 env 名是非法的 `ANTHROPIC-PROXY_PERSONAL_ACCESS_TOKEN`。其他两个分支（`valueAPIKey` / `valueBaseURL`）走的是 `lookupProviderEnv`，已含 dash → underscore 归一；此 PAT 分支独立拼接绕过了这一保护。

**复现**：用户 export `ANTHROPIC_PROXY_PERSONAL_ACCESS_TOKEN=xxx`，跑 `--model anthropic-proxy/foo`，`HasProvider("anthropic-proxy")` 在 PAT 分支返回 false（找的是不存在的 `ANTHROPIC-PROXY_...`），导致 collapse 触发，model 被当字面 id —— 与用户预期反着来。

**建议**：复用 `lookupProviderEnv` 风格做 dash → underscore 归一，或抽一个 `pattenvName(name string) string` helper 供这一处和 `lookupProviderEnv` 共用。

zpzjzj (commented):
## 第二轮评审报告（自动）

| 项目 | 内容 |
|------|------|
| **评审人** | @zpzjzj (via Claude Code github-code-review skill, monitor mode) |
| **HEAD** | `26239e02` (3 new commits since v1: `724ae51a`, `7f7a3c7e`, `26239e02`) |
| **规模** | 6 files, +648 / -9（新增 `internal/credential/provider_query.go` + test） |

### 第一轮意见处置（重点）

| # | 第一轮意见 | 状态 | 说明 |
|---|------------|------|------|
| 1 | 三处重复 split + ProviderConfigured | 🟡 **部分** | `ResolveModelRef` 收敛了回退逻辑；但全局仍 4 处 `SplitN`（`run.go:432`, `run.go:488`, `provider_query.go:91`, `agent_init.go:103`）|
| 2 | `ProviderConfigured` 类型泄漏 / 职责错位 | ✅ **已解决** | 改 `Resolver.HasProvider`，nil-safe |
| 3 | 静默回收无可观测性 / env fallback 无 log | ✅ **已解决** | collapse 路径补 `logging.Debugf`，覆盖测试齐备 |
| 4 | `--model` flag help string 不一致 | ✅ **已解决** | `run.go:103` 已改 |
| 5 | 可观测性快照早于决策 | ❌ **未解决** | 见下「Major #1」 |
| 6 | per-engine 稳定性顾虑 | 🟡 文档化为未来扩展点（acceptable） |
| 7 | CHANGELOG 未更新 | 🟡 **部分** | 0.1.2 加了 credentials.yaml 条目，但 4 个新 fix commit 全部未入 |
| 8 | cli-reference.md 描述失真 | 🟡 **部分** | `--api-key` 已补；`--model` 行仍写 "format: provider/name"，未反映新语义 |
| 9 | --model help string | ✅ |
| 10 | revert 历史代码注释 | ❌ **未解决** | grep `preserve provider prefix` / `revert` 在 .go 源码无任何注释 |
| 11 | user-config.md provider/credentials 入口 | ❌ **未解决** | 该文件 0 处提到 credentials.yaml、env 三件套、framework-default providers |

**整体进展**：方向正确，核心抽象（`ResolveModelRef` + `HasProvider`）已就位，重构与测试覆盖到位（202 行 test 覆盖 nil-receiver / framework-default / debug-log 等关键路径）。剩余架构债集中在「split 两阶段」与「快照 / 决策 phase 顺序」。

### 整体评价

**⚠️ 有条件合并** — 无 Blocker。建议本 PR 内修 v2 Major #1 #2（可观测性快照 + judge 侧 split 不对称），其余可作为 follow-up；文档 Major 必须补齐再合（特别是 CHANGELOG / cli-reference / user-config）。

CI 评审时 5 个 check 仍全 QUEUED（run #26218814647 + 后续 run），待跑完后回查。

### 行评论列表（已发布 v2）

| # | 文件:行 | 严重度 | 问题 |
|---|---------|--------|------|
| 1 | `cli/run.go:432` | 🟠 | `resolveModelRefWithCLIHints` 手写 split 是第 4 处 + hints 旁路 HasProvider 无 warn |
| 2 | `credential/provider_query.go:36` | 🟠 | `HasProvider` 硬编码 framework defaults + 函数语义双重身份 |
| 3 | `credential/provider_query.go:51` | 🟠 | PAT env 名拼接绕过 dash→underscore 归一（破折号 provider 名失效）|

### Major（PR 未触碰原行，无法挂行评论）

- **🟠 #1 可观测性快照仍早于 collapse 决策** — `internal/cli/run.go:187-194` vs `:235`：`loadAndPrepareConfig` 在 resolver 加载 + collapse 之前就把 `skill_up.model` 写进 span 并 `RecordRunStarted`。对 `--model unconfigured/foo` 走 collapse 后实际 model 是字面 `unconfigured/foo`，span attribute 已被记录为 `provider=unconfigured / name=foo` 拼回的字符串 —— 表面相同、维度切分错误，trace 上 provider 维度的分桶会污染。**建议把这段挪到 `loadCredentialsAndAgent` 之后，或在 collapse 后补一次 `span.SetAttributes` 覆盖**。

- **🟠 #2 判官模型仍走朴素 split，与 runner 语义不对称** — `internal/credential/agent_init.go:102-109`：runner 现在用 `ResolveModelRef` disambiguate；judge 的 `parseJudgeModel` 仍 `SplitN("/", 2)` 直接当 Provider/Name。同一字符串 `anthropic_modelscope/deepseek-v4-pro` 在 runner=collapse、judge=split，**两路径凭据解析行为分叉**。建议 `parseJudgeModel` 改委派 `ResolveModelRef`（透传 resolver 进 `ResolveJudgeInitParams`）。

- **🟠 #3 `resolveEvalConfig` 仍乐观 split** — `internal/cli/run.go:488-497`：在 resolver 加载前就把 `--model X/Y` 拆进 `evalCfg.Engine.Model.{Provider,Name}`，紧接着 `loadCredentialsAndAgent` 又调 `collapseUnconfiguredProviderSplit` 把可能错的拆分拼回去。「先猜后撤」反模式没消除，新增 collapse + helpers 整套都在补这个洞。**长期方案：把 `resolver.Load()` 提前到 Phase 1 末尾，删 `resolveEvalConfig` 里的手写 split 与 collapse helper，让 `ResolveModelRef` 成为唯一入口**。

- **🟠 #4 CHANGELOG 还缺 4 条** — `## [0.1.2] / ### Fixed` 应追加：(a) `--api-key` 现在优先于 env / credentials.yaml / Qoder PAT；(b) `eval.yaml engine.model.base_url` 不再被忽略；(c) anthropic / openai 视为 framework-default providers；(d) `--model provider/name` 在 provider 未配置时回退为字面 model id。

- **🟠 #5 `docs/guide/cli-reference.md --model` 行仍失真** — 当前写 "Override model (format: `provider/name`)"，未反映 ResolveModelRef 新语义。建议改为 "bare model name or `provider/name`（仅当 provider 已配置时拆分）"；`--api-key` 行补 precedence：`--api-key` > `<PROVIDER>_API_KEY` env > credentials.yaml > Qoder PAT。中文 `docs/zh/guide/cli-reference.md` 同步镜像。

### Info 级建议

代码：
- 测试 `provider_query_test.go:36-39` 直接写未导出 `r.creds["custom-provider"]`，耦合实现细节；用 `resolver.Load()` from yaml fixture 更稳
- `provider_query.go:10-30, 56-86` 与 `run.go:413-470` 5 个相邻函数各自带 10~30 行 doc，rationale 多次重复；建议抽到 `internal/credential/doc.go` 或 package-doc，函数 doc 只留契约

架构：
- `HasProvider` 是 `*Resolver` 方法、`ResolveModelRef` 是包级函数取 `*Resolver` 参数 —— 两者职责同源、调用约定不一致。建议改为 `(r *Resolver) ResolveModelRef(raw string)` 并保持 nil 安全
- 回归风险：`applyCLIOverrides` 流程对 `evalCfg.Engine.Model.Provider/Name` 的可变性多了一道「启动后再变一次」；单元测试覆盖到主要分支，但 integration 路径建议 `make test` + cli e2e 跑一遍

文档：
- `docs/zh/guide/{cli-reference,user-config}.md` 镜像未做（中文文件存在同款章节）
- `revert` 历史建议在 `collapseUnconfiguredProviderSplit` 或 claude_code agent 文件头加 "Historical note: an earlier agent-side preserve-prefix attempt (commit 632abc3) was reverted by b4f8f57 because the split happens before the resolver loads."，避免后人重蹈
- `user-config.md` 缺 "Credentials & Providers" 章节（引入 `~/.skill-up/credentials.yaml` 路径、三种 env 形式、framework-default providers、`engine.model.base_url` 与 provider 的关系）

### 已具备的好实践

- `ResolveModelRef` / `HasProvider` 抽象 + 命名清晰、godoc 详尽、测试覆盖到位
- collapse 路径 debug log + 测试覆盖（`TestResolveModelRef_DebugLogsOnCollapse` / `_NoLogOnConfiguredSplit`）双向
- 新增 3 commit 共 +463 行测试 / +194 行生产，**测试与变更意图直接对齐**（`--api-key` / `base_url` / framework-default 三 fix 各有专测）
- 决策点选址正确（CLI/credential 边界 vs agent 内部，比上一版 effectiveModelName 稳定）

---

**触发**：Monitor mode 在 push `26239e02` 后自动触发；按规则只评最新一次 push（v1 评的 `c670ff94`、中间的 `724ae51a` / `7f7a3c7e` 不重评）。本评审由 4 个并行子 Agent（代码质量、架构、CI、文档）完成；性能 / 安全因本 PR 无相关攻击面/热路径，跳过。

internal/cli/run.go
zpzjzj: Addressed in commit 2a9c7ce. `credential.ResolveModelRef` now takes a variadic `extraConfigured ...func(prefix string) bool` predicate; `resolveModelRefWithCLIHints` collapses to a 4-line wrapper that hands a `func(_ string) bool { return true }` predicate when `--api-key` or `evalBaseURL` is present, so the SplitN itself only lives in `ResolveModelRef` now. That kills the duplicated split site here and gives us the "engine-specific predicate" extension knob the godoc has been promising.

`resolveEvalConfig` still has its own SplitN (it runs pre-resolver) and `parseJudgeModel` is its own judge-side path — those remain the wider refactor flagged in [[第一轮 #1+#2]]; this commit just closes the in-scope duplicate.

On the silent-accept-typo'd-provider concern: the CLI-hint path still upgrades any prefix to "configured" without a Warn, identical to pre-2a9c7ce behavior — not a regression, but worth a follow-up. Tracked separately so the diff here stays scoped.

internal/credential/provider_query.go
zpzjzj: Partially addressed in commit 2a9c7ce:

1. **Framework-defaults inline → named extraction**: pulled into package-private `frameworkDefaultProviders` slice + `isFrameworkDefaultProvider` helper. Rationale now lives next to the data; adding a new engine default is a one-line append.
2. **Dual-identity doc**: `HasProvider`'s godoc now explicitly calls out the "treat-as-namespace vs credentials-exist" caveat and points strict callers at `Resolver.Get + lookupProviderEnv`.
3. **`frameworkDefaults` as a `Resolver` field + rename to `ShouldKeepProviderSplit` + `reconcileProviderSplit`**: deferred. Each is a wider rename that ripples through tests and call sites; doing them now would balloon the diff past the PR's scope. The caveat doc captures the dual semantics so future readers aren't surprised, and the extraction lets the rename happen mechanically in a follow-up.

internal/credential/provider_query.go
zpzjzj: Premise check turned up a different picture than the v2 note assumed. `lookupProviderEnv` (`internal/credential/agent_init.go:225-235`) does **not** normalize dashes — it's a straight `strings.ToUpper(provider) + "_" + kind`. So the `valueAPIKey` / `valueBaseURL` branches and this PAT branch behave identically for dash-bearing provider names: both produce `ANTHROPIC-PROXY_*`-shaped env names, and neither is reachable in standard POSIX shells.

There's no inconsistency between branches to fix here. The deeper "dash provider names don't round-trip through env" issue is a pre-existing limitation that predates this PR — solving it means normalizing in `lookupProviderEnv` (and matching in every caller that exports envs by hand). Leaving as-is in this PR; happy to do the normalization pass as a follow-up if dash-bearing provider names become a supported configuration.

internal/cli/run.go
```diff
@@ -401,6 +411,36 @@ func evaluateOptionsFromFlags(cmd *cobra.Command) (runner.EvaluateOptions, error
...
+	provider := evalCfg.Engine.Model.Provider
+	name := evalCfg.Engine.Model.Name
+	if provider == "" || name == "" {
+		return
+	}
+	if credential.ProviderConfigured(provider, resolver) {
+		return
+	}
+	evalCfg.Engine.Model.Provider = ""
+	evalCfg.Engine.Model.Name = provider + "/" + name
```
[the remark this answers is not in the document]
zpzjzj: Addressed in commits 724ae51 + 26239e0. `collapseUnconfiguredProviderSplit` and the sibling `normalizeCLIModelOverride` now both go through `resolveModelRefWithCLIHints`, which honours `cliAPIKey` as a "configured" signal — when `--api-key` is the only credential source the prefix is preserved as `Provider`, so `applyCLIOverrides` still applies the key. Regression-covered by `TestNormalizeCLIModelOverride_CLIAPIKeyKeepsSplitStripping`.

internal/credential/agent_init.go
```diff
@@ -234,6 +234,41 @@ func lookupProviderEnv(provider string, kind scopedValueKind) (value, envVar str
...
+	if resolver != nil {
+		if _, ok := resolver.Get(provider); ok {
+			return true
+		}
+	}
+	if _, _, ok := lookupProviderEnv(provider, valueAPIKey); ok {
+		return true
+	}
+	if _, _, ok := lookupProviderEnv(provider, valueBaseURL); ok {
+		return true
```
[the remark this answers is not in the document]
zpzjzj: Addressed in commit 7f7a3c7. `Resolver.HasProvider` now probes `<PROVIDER>_PERSONAL_ACCESS_TOKEN` alongside the API_KEY / BASE_URL envs, so a Qoder setup with only `QODER_PERSONAL_ACCESS_TOKEN` exported registers as configured and `--model qoder/auto` keeps its split. Covered by `TestHasProvider_TrueWhenPersonalAccessTokenEnvSet`.

internal/cli/run.go
```diff
@@ -224,11 +224,21 @@ func loadCredentialsAndAgent(cmd *cobra.Command, evalCfg *config.EvalConfig) (ag
...
 
+	// `--model provider/name` was tentatively split into Provider+Name by
+	// resolveEvalConfig (before the resolver was loaded). Now that we know
+	// which providers actually have configuration, collapse the split back
+	// when the provider is unknown: in that case the slashed string is more
+	// likely a literal model identifier the upstream API expects verbatim
+	// (e.g. anthropic-proxy gateways registering models under
+	// `anthropic_modelscope/deepseek-v4-pro`) than a credential-namespace
+	// prefix the agent should peel off.
+	collapseUnconfiguredProviderSplit(evalCfg, resolver)
```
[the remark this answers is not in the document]
zpzjzj: Addressed in commit 26239e0. `Resolver.HasProvider` now returns true unconditionally for the framework-default providers (`anthropic`, `openai`, case-insensitive). Their upstream APIs only accept bare model ids and the underlying CLIs (claude, codex) carry persistent login state the resolver can't see — so `--model anthropic/X` with login-state-only configuration no longer collapses. Covered by `TestHasProvider_TrueForFrameworkDefaults` and `TestCollapseUnconfiguredProviderSplit_KeepsAnthropicSplitWithoutEnv`.

internal/cli/run.go
```diff
@@ -401,6 +411,61 @@ func evaluateOptionsFromFlags(cmd *cobra.Command) (runner.EvaluateOptions, error
...
+//
+// Either signal upgrades any prefix to "configured", regardless of
+// resolver/env state. Wired through ResolveModelRef's extraConfigured
+// predicate so the SplitN / collapse logic stays in one place.
+func resolveModelRefWithCLIHints(modelFlag string, resolver *credential.Resolver, cliAPIKey, evalBaseURL string) (provider, name string) {
+	if cliAPIKey == "" && evalBaseURL == "" {
+		return credential.ResolveModelRef(modelFlag, resolver)
+	}
+	cliHintConfigured := func(_ string) bool { return true }
+	return credential.ResolveModelRef(modelFlag, resolver, cliHintConfigured)
```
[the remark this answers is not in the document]
zpzjzj: Acknowledged but design-intentional. This is the same behavior as before commit 2a9c7ce — the prior `resolveModelRefWithCLIHints` did the same `len(parts)==2 → return parts[0], parts[1]` short-circuit when either hint was set, so this isn't a regression from the refactor. The rationale lives in the godoc above the function and on `applyCLIOverrides`:

- `--api-key K` requires a non-empty `Provider` to apply (`credential.applyCLIOverrides` early-returns otherwise). Treating `--api-key K --model anthropic_modelscope/deepseek-v4-pro` as opaque would drop K and the run would fail auth — strictly worse than the current "split and trust the user-supplied namespace" choice.
- `engine.model.base_url` in eval.yaml is a deliberate provider-configuration signal; the gateway URL is bound to the namespace.

The genuinely conflicting case (`--api-key K --model literal_opaque_id/Y`) has no resolution that satisfies both intents simultaneously without a new dedicated flag. The pragmatic choice is to preserve the split when CLI hints are present and document the edge case.

Adding a Warn log when the CLI-hint path upgrades a prefix that `HasProvider` would otherwise reject is a reasonable follow-up — the v2 self-review thread on this same function flagged it, and I'd treat both as one follow-up rather than expanding this PR's scope. Open to doing it next if you want to chat first.

internal/cli/run.go
```diff
@@ -401,6 +411,62 @@ func evaluateOptionsFromFlags(cmd *cobra.Command) (runner.EvaluateOptions, error
...
```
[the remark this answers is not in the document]
zpzjzj: Same point as the prior P1 on this same function — see [the resolved thread on root comment 3280924815](https://github.com/alibaba/skill-up/pull/42#discussion_r3280924815). Short version:

- Behavior is unchanged from pre-2a9c7ce (the previous `if cliAPIKey != "" || evalBaseURL != "" { … return parts[0], parts[1] }` did the same forced split).
- `--api-key K --model X/Y` requires non-empty `Provider` for `applyCLIOverrides` to apply K; treating the slashed id as opaque would silently drop the CLI key and fail auth — strictly worse.
- The `--api-key K --model literal_opaque_id/Y` case has no resolution that satisfies both intents without a dedicated flag.

Adding a Warn log on the CLI-hint upgrade path (when `HasProvider` would otherwise reject) is the right follow-up; tracked alongside the v2 self-review thread on this same function. Resolving as no-change-needed-in-this-PR.

zpzjzj:
Correction on the two earlier P1 threads I closed as "no-change-needed-in-this-PR" ([3280924815](https://github.com/alibaba/skill-up/pull/42#discussion_r3280924815), [3281096790](https://github.com/alibaba/skill-up/pull/42#discussion_r3281096790)):

**Codex was right and my closing rationale was wrong.** I claimed the `--api-key K + literal_opaque/id` case had no resolution that satisfies both intents without a dedicated flag — but that argument hinged on `applyCLIOverrides` requiring non-empty `Provider` to apply `K`. Re-reading the code, that guard was **defensive paranoia, not a structural requirement**: each agent routes `cfg.APIKey` through its own hardcoded env (`claude_code` → `ANTHROPIC_API_KEY`, `codex` → `OPENAI_API_KEY`, see `BaseAgent.credentialEnvVars`), so the key reaches upstream correctly regardless of `Provider`.

Fixed in commit 1836d12:

- `credential.applyCLIOverrides`: removed the `Provider==""` guard; `--api-key` applies unconditionally.
- `cli.resolveModelRefWithCLIHints`: deleted. CLI-hint signals no longer force a split — `ResolveModelRef` is the single decision point.
- `cli.collapseUnconfiguredProviderSplit` / `cli.normalizeCLIModelOverride`: drop their `cliAPIKey` / `evalBaseURL` parameters.
- `credential.ResolveModelRef`: drop the variadic `extraConfigured` parameter that existed solely for the CLI-hint path.

Net effect: `--api-key K --model anthropic_modelscope/deepseek-v4-pro` now sends `ANTHROPIC_API_KEY=K + --model anthropic_modelscope/deepseek-v4-pro` upstream as the user expects. Users wanting namespace semantics for unknown providers should configure that provider via env / credentials.yaml; `HasProvider` then preserves the split.

Test flip:
- `TestResolveRunnerInitParams_CLIAPIKeyAppliesWithoutProvider` (was `_WarnsWhenCLIAPIKeyHasNoProvider`)
- `TestCollapseUnconfiguredProviderSplit_PreservesSplit` drops the `CliAPIKeyProvided` and `EvalBaseURLProvided` cases (they now collapse — which is the correct behavior)

hittyt (approved):
LGTM

PR #58, opened 2026-05-22
fix(runner): generate report.json when --format json is specified for run subcommand

### LINKED ISSUE #57: `report.json` isn't generated when set report json config
## Description

When using `skill-up run` with `--format json` (or `formats: [json]` in `eval.yaml`), no `report.json` file is generated. The `"json"` case in `writeFormattedReports` is silently skipped with the comment "result.json already written", but `result.json` is always written regardless of format flags, so specifying `--format json` has no observable effect.

This also creates an inconsistency with `skill-up report`, where `--format json` **does** produce a `report.json` file.

## Steps to reproduce

1. Create a valid eval config with `formats: [json]` in `eval.yaml`, or run with `--format json`:
   ```bash
   skill-up run ./evals/eval.yaml --format json
   ```
2. Check the output iteration directory for generated files.
3. Observe that only `result.json` exists — there is no `report.json`.

## Expected behavior

Setting `--format json` (or `formats: [json]`) should produce a `report.json` file in the iteration directory, consistent with how `skill-up report --format json` generates `report.json`.

At minimum, the three format values should produce symmetrically named outputs:

| Format   | Expected file   |
|----------|-----------------|
| `json`   | `report.json`   |
| `junit`  | `report.xml`    |
| `html`   | `report.html`   |

## Actual behavior

The `"json"` branch in `writeFormattedReports` (`internal/runner/runner.go:317`) is a no-op:

```go
case "json":
    // result.json already written; skip.
```

- `result.json` is always written unconditionally (line 216), regardless of `--format`.
- `--format json` silently does nothing extra — no `report.json` is created.
- `--format junit` → `report.xml` ✅
- `--format html` → `report.html` ✅
- `skill-up report --format json` → `report.json` ✅ (inconsistent with `run`)

## Environment

- skill-up version: dev (main branch)
- Go version: 1.25+
- OS / arch: all

## Additional context

Relevant code locations:

- **`internal/runner/runner.go`** `writeFormattedReports` (line ~312): the `"json"` case is a no-op skip.
- **`internal/cli/report.go`** `makeReporter` (line ~97): the `"json"` case correctly creates `report.json`.

A simple fix would be to have the `"json"` case in `writeFormattedReports` write `report.json` via `JSONReporter`, mirroring how `junit` and `html` are handled, while keeping `result.json` as the always-present raw data file.

roark47:
## Summary

When using `skill-up run --format json` (or `formats: [json]` in eval.yaml), no `report.json` file was generated. The `"json"` case in `writeFormattedReports` was silently skipped with the assumption that `result.json` (always written unconditionally) was sufficient. This created an inconsistency with `skill-up report --format json`, which correctly produces `report.json`.

This PR makes the `run` subcommand also emit `report.json` when `--format json` is requested, aligning behavior across both entry points.

## Related issues

Closes #57 

## Changes

- **`internal/runner/runner.go`**: Updated `writeFormattedReports` to write `report.json` via `JSONReporter` instead of silently skipping the `"json"` format. `result.json` continues to be written unconditionally as the raw data source.

## Test plan

- [x] `make test` passes
- [ ] Manual testing steps (if applicable):
  1. Run `skill-up run ./evals/eval.yaml --format json`
  2. Verify `report.json` exists in the iteration output directory
  3. Verify `result.json` is still present (unchanged behavior)

## Notes for reviewers

- `result.json` is still always written regardless of `--format` flags — this is intentional and unchanged.
- The new `report.json` is identical in content to `result.json`; the difference is that it is only produced when explicitly requested via `--format json`.
- This aligns `skill-up run` with `skill-up report`, where `--format json` already produces `report.json` (see `internal/cli/report.go` line 97).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.2.1] - 2026-05-22
+
+### Fixed
+- `skill-up run --format json` now generates `report.json` in the iteration
+  output directory. Previously the `"json"` format was silently skipped
+  because `result.json` is always written unconditionally; this made
+  `--format json` a no-op and was inconsistent with `skill-up report --format json`
+  which correctly produced `report.json`.
+
 ## [0.2.0] - 2026-05-21
 
 ### Changed
@@ -98,6 +107,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   project and delivers the end-to-end capability to declare eval environments,
   run cases and emit structured reports as described in [README.md](README.md).
 
+[0.2.1]: https://github.com/alibaba/skill-up/releases/tag/v0.2.1
 [0.2.0]: https://github.com/alibaba/skill-up/releases/tag/v0.2.0
 [0.1.2]: https://github.com/alibaba/skill-up/releases/tag/v0.1.2
 [0.1.1]: https://github.com/alibaba/skill-up/releases/tag/v0.1.1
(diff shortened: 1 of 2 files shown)


zpzjzj (approved):
LGTM

lijunfeng722 (approved):
LGTM

PR #60, opened 2026-05-25
fix(config): apply documented defaults when loading eval.yaml

zpzjzj:
## Summary

- `LoadEvalConfig` previously unmarshalled `eval.yaml` into a zero-value struct, so the implicit defaults documented in `README.md:167` and `docs/guide/getting-started.md:140` (`timeout_seconds: 300`, `max_turns: 10`, `parallelism: 1`, `report.formats: [json]`) only applied on the *no-eval.yaml* code path through `DefaultEvalConfig()`.
- With an `eval.yaml` present, an omitted `timeout_seconds` resolved to `0` and `withCaseTimeout` (`internal/evaluator/evaluator.go:661`) treats `<= 0` as **no deadline** — the opposite of what the docs promise.
- Now apply those four documented defaults after unmarshalling, sourcing values from `DefaultEvalConfig()` so `defaults.yaml` stays the single source of truth.

## Why this scope (only 4 fields)

`DefaultEvalConfig()` also has values for `schema_version: v1alpha1`, `environment.type: none`, `engine.name: claude_code`, but those are *required* per `internal/config/validator.go:34-53`. Layering the full defaults onto user input would silently mask documents that forgot to declare them — a regression in validation. So this PR only fills the four fields the documentation actually advertises as having defaults.

## Semantics: `timeout_seconds: 0`

To keep the "use default if unset" rule simple, `timeout_seconds: 0` is treated as *not configured* and resolves to 300. Users who genuinely want no deadline can pass `-1` (or any negative value) — `withCaseTimeout` already treats `<= 0` as "no deadline". A regression test for the `-1` form is included.

## Real-world trigger

Came up while diagnosing a CI run where one `update-user-name` case timed out at ~3 min in `agent_judge`. The author had not set `timeout_seconds` themselves and reasonably expected the documented 300s default — but no deadline was being applied at all, so the actual budget was implicit/unbounded until something else interrupted.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (all packages green, including `internal/cli` which exercises `loadFromEvalYAML` + `applyImplicitSkills`)
- [x] New table-driven `TestLoader_LoadEvalConfig_AppliesDefaults` covers: omitted `cases.defaults`, partial `cases.defaults`, explicit override, `0` → falls back to default, `-1` → no deadline, user `report.formats`, user `workspace_mount` (not auto-defaulted).
- [x] `TestLoadFromEvalYAML_AddsImplicitLocalSkillWhenSkillsOmitted` and `TestLoadFromEvalYAML_RespectsExplicitEmptySkills` still pass (the `Skills==nil` vs `Skills==[]` distinction is preserved because we don't touch `Skills`).
- [x] `gofmt`, `go vet` clean.

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -57,7 +57,11 @@ func NewLoader(evalPath string) *Loader {
 	}
 }
 
-// LoadEvalConfig loads the eval.yaml file.
+// LoadEvalConfig loads the eval.yaml file and fills in the documented
+// implicit defaults (timeout_seconds, max_turns, parallelism, report.formats)
+// for fields the user omitted. Other unset fields keep their zero value so the
+// validator can still reject documents that omit required keys like
+// schema_version, environment.type, and engine.name.
 func (l *Loader) LoadEvalConfig() (*EvalConfig, error) {
 	data, err := os.ReadFile(l.evalPath)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


roark47 (approved):
LGTM

PR #63, opened 2026-05-25
refactor(agent,runtime): unify PATH/env handling via command prelude

zpzjzj:
## Why

The agent layer was injecting a shell-expansion string —
`$HOME/.local/bin:$HOME/.nvm/current/bin:$PATH` — as the literal value of
the `PATH` key in `map[string]string` env tables. Process-level env vars
are not shell-expressions, so every runtime had to compensate:

- **NoneRuntime** ran user-provided values through `os.Expand` against
  the host environment to coincidentally produce the right PATH.
- **DockerRuntime** and **OpenSandboxRuntime** each detected `PATH`
  values containing `$`, stripped them from the literal env list, and
  prepended an `export PATH="…"` line to the command so the inner
  shell would expand them. Docker additionally rejected `$(…)` to
  guard against command substitution firing inside that auto-generated
  export.

Three structurally identical patches lived in three files (~23 lines of
defensive commentary in `docker.go` alone), plus `NoneRuntime` behaved
differently from the other two for user-supplied `environment.env`
values.

## What

Move PATH from the env-map layer to the command layer, where shell
expansion lives.

1. **Agent layer** — new `withAgentExecPath(cmd)` helper prepends
   `export PATH="$HOME/.local/bin:$HOME/.nvm/current/bin:$PATH"` to the
   command itself. `mergeExecOptionsEnv` no longer injects `PATH` into
   the env map. Applied to all 8 agent Install / Run / Check call sites
   (claude-code, codex, qodercli, generic CLI).
2. **Runtimes** — docker and opensandbox now pass every env var
   literally; no key gets special handling. The `pathPrefix` branch,
   the `$(…)` rejection, `shellDoubleQuote`, `literalRemoteEnv`,
   `withRemoteEnvExpansion`, and the matching cfg-time PATH skip are
   all deleted.
3. **NoneRuntime** — `mergeEnv` no longer routes values through
   `expandEnvMap` / `os.Expand`. Values forward literally, matching
   docker and opensandbox. This is the user-visible breaking change
   (see CHANGELOG).

Commits are atomic, in the order an audit will read them:

```
9d12711 refactor(agent): move PATH injection from env map to command prelude
479ff32 refactor(runtime): drop PATH special-case in docker/opensandbox Exec
bece2ef refactor(runtime): drop NoneRuntime env expansion for consistency
226344a docs(e2e): update mockEngineHome comment to reflect prelude mechanism
```

## Safety / attack surface

The deleted `$(…)` rejection in `docker.go` guarded an attack vector
that only existed because the runtime was constructing
`export PATH="<user-value>"` from caller-supplied env. With env passed
via `--env KEY=VALUE` / sandbox `Envs` (process-level env-var setting,
no shell parsing), that vector no longer exists. The agent layer's
`withAgentExecPath` uses a compile-time constant (`agentExecutablePath`),
never user input.

## Breaking change

`NoneRuntime` no longer expands `\$VAR` / `\${VAR}` in
`environment.env` or `opts.Env` values. Configs that depended on
host-side expansion (e.g. `MY_VAR: \"\$HOME/foo\"`) should switch to
literal values or move the export into a `setup_steps` shell snippet.
Documented in CHANGELOG.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (all packages green)
- [x] `go test -tags=e2e ./e2e/... -run TestPipeline` — end-to-end mock-engine path, which exercises the agent → runtime PATH prelude chain
- [x] Existing PATH-injection assertions in `agent_test.go` / `claude_code_test.go` / `codex_test.go` / `qodercli_test.go` flipped to negative: env map must NOT carry PATH, command must start with the export prelude
- [x] New `TestDockerRuntime_ExecPassesEnvLiterally` confirms PATH and `\$(…)`-containing values forward literally
- [x] New `TestNoneRuntime_ForwardsEnvLiterally` confirms no host-side expansion
- [ ] CI to confirm on Linux runners

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,32 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- `runtime.Runtime.MergeEnv(env)`: new interface method that lets
+  setup-time callers (notably each agent's `Install`) seed the runtime's
+  persistent env baseline. Subsequent `Exec` calls inherit these vars
+  unless overridden by `opts.Env`. Implementations are intended for
+  one-time setup; concurrent `MergeEnv`/`Exec` is not safe.
+
+### Changed
+- Agent layer no longer injects PATH into the env map. Each agent's
+  `Install` now probes the target shell for the resolved PATH (a
+  per-agent `printf '%s' "$HOME/..."` command) and merges the literal
+  result into the runtime's env baseline via `runtime.MergeEnv`.
+  Runtimes treat env keys uniformly — no key is special-cased anywhere.
+- `NoneRuntime` no longer expands `$VAR` / `${VAR}` in user-provided env
+  values (`environment.env` / `opts.Env`). Values forward literally,
+  matching docker and opensandbox. If your `eval.yaml` relied on
+  host-side expansion (e.g. `MY_VAR: "$HOME/foo"`), switch to a literal
+  value or `export` it inside a `setup_steps` step.
+- `environment.type: none`: the framework no longer force-prepends
+  `$HOME/.local/bin:$HOME/.nvm/current/bin:$PATH` to agent commands.
+  Because `ag.Install` is skipped for type=none, the new probe doesn't
+  run either; the host shell's PATH is used as-is. Add the dirs to
+  your shell rc if `claude` / `codex` / `qodercli` isn't already on it.
+
 ## [0.2.3] - 2026-05-27
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   matching docker and opensandbox. If your `eval.yaml` relied on
   host-side expansion (e.g. `MY_VAR: "$HOME/foo"`), switch to a literal
   value or `export` it inside a `setup_steps` step.
+- `docker` / `opensandbox`: the same "values forward literally" rule
+  now applies — previously the docker runtime silently dropped
+  `environment.env.PATH` values containing `$` (the alternative was
+  passing them to `docker create --env` literally, which would have
+  broken container startup because the entrypoint's PATH lookup can't
+  resolve directories named `$HOME` / `$PATH`). After this PR no key
+  is special-cased, so passing such a value will now make the
+  container/sandbox fail to start; use a literal PATH or move the
+  manipulation into a `setup_steps` `export`.
 - `environment.type: none`: the framework no longer force-prepends
   `$HOME/.local/bin:$HOME/.nvm/current/bin:$PATH` to agent commands.
   Because `ag.Install` is skipped for type=none, the new probe doesn't
--- a/internal/runtime/docker.go
+++ b/internal/runtime/docker.go
@@ -166,9 +166,6 @@ func (r *DockerRuntime) buildCreateArgs(name string) []string {
 		args = append(args, "--network", "none")
 	}
 	for k, v := range r.cfg.Env {
-		if k == "PATH" && strings.Contains(v, "$") {
-			continue
-		}
 		args = append(args, "--env", k+"="+v)
 	}
 	entry := r.cfg.Entrypoint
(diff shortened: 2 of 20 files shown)


zpzjzj:
@codex review

internal/runtime/docker.go
```diff
@@ -165,9 +165,6 @@ func (r *DockerRuntime) buildCreateArgs(name string) []string {
 		args = append(args, "--network", "none")
 	}
 	for k, v := range r.cfg.Env {
-		if k == "PATH" && strings.Contains(v, "$") {
-			continue
-		}
 		args = append(args, "--env", k+"="+v)
```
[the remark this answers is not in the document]
zpzjzj: Confirmed and fixed in 5378084.

You're right — execvpe on the container side searches the literal `$HOME` / `$PATH` directory names, finds nothing, and the entrypoint (default `sleep infinity`) never starts. The same lookup hazard applies to `docker exec ... sh -c "<cmd>"` (looking up `sh`) and to opensandbox's own shell-binary lookup.

The earlier cleanup correctly removed the runtime-side `export PATH=…` auto-construction (and the `$(…)` rejection that only guarded that construction), but it should not have removed the `cfg.Env` / `opts.Env` PATH-with-`\$` skip itself — that guard exists for a different reason (protecting the runtime's own process lookup), not for the shell-construction path.

Restored the skip at all three call sites: `buildCreateArgs` in docker, `Exec` in docker, and `Exec` in opensandbox. Added a warning log so users get feedback instead of a silent drop. The agent's PATH continues to flow through the command-side `export PATH=…` prelude (unaffected). Tests added — `TestDockerRuntime_CreateDropsPATHWithShellExpansionFromCfgEnv` covers the create-time path explicitly — and CHANGELOG updated.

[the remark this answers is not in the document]
zpzjzj: Reopening this — on reflection the previous "fix" (commit 5378084, now force-pushed away) was treating a non-problem. Reverted: the runtime no longer special-cases PATH-with-`$` in `cfg.Env` / `opts.Env`.

Why:

1. **No real user hits this.** Repo-wide grep across `examples/` and `skills/` finds zero `eval.yaml` setting `environment.env.PATH` at all (the one `env:` block sets `TZ: UTC`). The "evals that set `environment.env.PATH` using shell references" scenario is hypothetical.

2. **It's inconsistent with the new design.** This PR's CHANGELOG already documents the breaking change "env values are passed literally; if you want shell expansion, prepend `export X=...` to the command". PATH is just a specific case of that rule — singling it out for runtime-side drop while every other key passes literally is dressing up.

3. **The protection is partial anyway.** It only catches `$`-containing PATHs. A user writing literal-but-incompatible `PATH: "/wrong:/path"` (no `/bin`) would also fail `sh` lookup, and the `$` check wouldn't save them. Half-fixing one foot-gun pattern while leaving the same class of failure open is worse than just documenting the rule once.

4. **The failure mode is loud, not silent.** A real user who writes `PATH=$HOME/...:$PATH` sees `executable file not found in $PATH` at container startup — clear enough to point at the cause. The previous "silently drop and pretend it took effect" behavior was arguably worse UX.

Net change: runtime is now fully uniform — no key gets special treatment — which is what the cleanup PR was about in the first place. The previously-dropped 5378084 had grown a 3-site duplication that on review felt like protecting a phantom user.

Happy to revisit if you can point to a concrete eval.yaml that depends on this, or to a failure mode beyond the loud "executable file not found" startup error.

CHANGELOG.md
```diff
@@ -5,6 +5,41 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
roark47: Need specific version here

zpzjzj: shall be changed during release changelog pr

zpzjzj: Done in 9cd22a3 — bumped to `[0.2.4] - 2026-05-27` (next minor after 0.2.3). Happy to switch to 0.3.0 if you'd rather treat the `runtime.MergeEnv` interface widening + NoneRuntime expansion drop as major. Adjust the date at merge time if it slips.

zpzjzj: Pushing back gently — could you confirm the intent here? Two reasons I'd default to keeping `[Unreleased]`:

1. **Keep a Changelog convention** — the file's header (line 5) explicitly says "based on Keep a Changelog", and KaC's spec says: *"There should be an 'Unreleased' section at the top to track upcoming changes."* The version + date go in only at release time.

2. **This repo's own practice** — the last two releases (0.2.2 in commit b0d1b14, 0.2.3 in 84b8919) were both stamped by you in dedicated `docs(changelog): record 0.2.x ...` commits at release time, not by the original PR authors. PR authors landing into 0.2.2/0.2.3 wrote `[Unreleased]` and you renamed it on release.

Picking a version myself in the PR risks (a) collision if multiple PRs all claim 0.2.4, (b) the date being wrong by the time of merge, (c) over-assigning a minor when you might prefer to ship a 0.3.0 for the breaking changes here.

I reverted the version-bump commit (force-pushed). If you'd like me to assign anyway — happy to, just confirm which version (0.2.4 vs 0.3.0) and I'll re-stamp.

roark47 (approved):
LGTM

PR #71, opened 2026-05-26
test: improve unit coverage

roark47:
## Summary
- Add meaningful unit coverage across CLI command paths, evaluator lifecycle helpers, MCP provisioning/install behavior, OpenSandbox helpers, observability, reporting, user config, UI, shell quoting, and public wrapper/transcript APIs
- Cover dry-run/config loading, artifact download behavior, environment/reference validation, metrics initialization, and path-safety boundaries
- Raise aggregate unit coverage above 85%

## Verification
- GOCACHE=/private/tmp/skill-up-gocache make fmt
- GOCACHE=/private/tmp/skill-up-gocache make verify
- GOCACHE=/private/tmp/skill-up-gocache make test
- GOCACHE=/private/tmp/skill-up-gocache go test -coverprofile=/tmp/skill-up-cover-final.out ./... -> 85.02%

### DIFF
--- a/internal/cli/command_integration_test.go
+++ b/internal/cli/command_integration_test.go
@@ -28,6 +28,16 @@ func captureStdout(t *testing.T, fn func() error) (string, error) {
 		done <- copyErr
 	}()
 
+	cleaned := false
+	defer func() {
+		if cleaned {
+			return
+		}
+		_ = w.Close()
+		os.Stdout = orig
+		_ = r.Close()
+	}()
+
 	runErr := fn()
 	_ = w.Close()
 	os.Stdout = orig
@@ -35,6 +45,7 @@ func captureStdout(t *testing.T, fn func() error) (string, error) {
 		t.Fatalf("copy stdout: %v", copyErr)
 	}
 	_ = r.Close()
+	cleaned = true
 
 	return buf.String(), runErr
 }
--- a/internal/cli/command_integration_test.go
+++ b/internal/cli/command_integration_test.go
@@ -1,0 +1,91 @@
+package cli
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/spf13/cobra"
+)
+
+func captureStdout(t *testing.T, fn func() error) (string, error) {
+	t.Helper()
+
+	orig := os.Stdout
+	r, w, err := os.Pipe()
+	if err != nil {
+		t.Fatalf("pipe stdout: %v", err)
+	}
+	os.Stdout = w
+
+	var buf bytes.Buffer
+	done := make(chan error, 1)
+	go func() {
+		_, copyErr := io.Copy(&buf, r)
+		done <- copyErr
+	}()
+
+	runErr := fn()
+	_ = w.Close()
+	os.Stdout = orig
+	if copyErr := <-done; copyErr != nil {
+		t.Fatalf("copy stdout: %v", copyErr)
+	}
+	_ = r.Close()
+
+	return buf.String(), runErr
+}
+
+func TestValidateCommandRunEReportsLoadedCases(t *testing.T) {
+	cmd := &cobra.Command{}
+	cmd.SetContext(context.Background())
+
+	output, err := captureStdout(t, func() error {
+		return validateCmd.RunE(cmd, []string{testEvalPath})
+	})
+	if err != nil {
+		t.Fatalf("validate RunE returned error: %v", err)
+	}
+	if !strings.Contains(output, "eval.yaml is valid") || !strings.Contains(output, "loaded 3 case(s)") {
+		t.Fatalf("validate output = %q, want success with case count", output)
+	}
+}
+
+func TestValidateCommandRunEWrapsLoadErrors(t *testing.T) {
+	cmd := &cobra.Command{}
+	cmd.SetContext(context.Background())
+
+	_, err := captureStdout(t, func() error {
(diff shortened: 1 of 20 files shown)


lijunfeng722:
@codex review

internal/cli/command_integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+	var buf bytes.Buffer
+	done := make(chan error, 1)
+	go func() {
+		_, copyErr := io.Copy(&buf, r)
+		done <- copyErr
+	}()
+
+	runErr := fn()
+	_ = w.Close()
+	os.Stdout = orig
```
zpzjzj, line 43: **🟠 Major: `captureStdout` teardown is not deferred — `fn()` panic leaks `os.Stdout` and an `io.Copy` goroutine**

The restore (`os.Stdout = orig`, `w.Close()`, `r.Close()`) runs only on the happy path, after `fn()` returns normally. If `fn()` panics (or any future change makes it possible to), the following happens:

- `os.Stdout` stays pointing at the pipe writer `w`.
- `w.Close()` is never called, so the `io.Copy(&buf, r)` goroutine never sees EOF and blocks forever.
- Every subsequent test that writes to stdout sends bytes into that dead pipe; they get captured into an unreferenced `buf` and disappear from CI logs.

The helper is reused by 5 tests across `command_integration_test.go`, `root_test.go:39`, and `run_test.go:99` (`runEval` path, the most complex of them). Today no caller is known to panic, but the moment one does the failure mode is silent and very hard to bisect.

**Suggested fix** — move the teardown into a `defer` so it runs on panic too:

```go
defer func() {
    _ = w.Close()
    os.Stdout = orig
    _ = r.Close()
}()
```

The `<-done` read can stay where it is — once `w.Close()` runs in the defer, the goroutine receives EOF and exits cleanly.

---
*Source: pr-review-curator skill @ 111d85a, filtered per docs/review-heuristics.md*

zpzjzj (commented):
## Code Review Report

| Item | Value |
|------|-------|
| **Review source** | pr-review-curator skill @ `111d85a`, filtered per docs/review-heuristics.md |
| **Scope** | Test-only PR (20 test files, +1810/-12, no production code touched) |
| **CI** | ✅ All 12 checks green |

### Finding summary

- 🔴 Blocker: 0
- 🟠 Major: 1 (posted as inline comment)
- 🔵 Info: 1 (below)

### Overall recommendation

✅ **Ship it.** No blockers. The one Major is a test-infrastructure robustness issue and can be fixed in a follow-up just as easily as here.

The new tests are high quality overall:

- Path-safety coverage (`workspace_test.go` `validateCaseID` / `WriteFile`, `opensandbox_test.go` `safeLocalTarget` / `remoteRelativePath`) is genuinely useful — it locks down the `..` / `/abs` / `nested\case` rejection behavior that is exactly the class of bug worth a regression test.
- Global-state mutations (`ui.Output`, `envConfigCache`, `instruments`, `createOpenSandbox`) are all paired with `t.Cleanup` restores, and the tests that touch globals correctly **do not** call `t.Parallel()`.
- Heavy use of `t.Setenv` / `t.TempDir` / `t.Chdir` — no leftover state, no environment pollution.
- Table-driven sub-tests use `t.Parallel()` where safe.
- No `time.Sleep`-based waits, no real network, no flaky timing assertions.
- Defensive-copy semantics verified explicitly in `TestContextAttributeAccessorsReturnCopies` — nice touch.

### Info-level note

**`TestSerializeTranscriptWritesJSONAndCleanupRemovesFile`** (`internal/evaluator/evaluator_test.go:261`) — the returned `cleanup` is only invoked on the success path. If either `os.ReadFile` or `strings.Contains` triggers `t.Fatalf`, the temp file is leaked. Consider `t.Cleanup(cleanup)` immediately after construction, then a separate explicit-cleanup assertion at the end. Cosmetic — not blocking.

### Filtering notes

Around 10 candidate findings were dropped during curation, including:

- "`TestLoadCredentialsAndAgentAppliesCLIModelAndAPIKey` should also unset `XDG_CONFIG_HOME`" — false alarm: the test pre-injects `userconfig.Config{}` via `userconfig.WithContext`, bypassing the loader's XDG branch entirely. Credential lookup uses `os.UserHomeDir()`, which the `HOME` override already covers.
- "`TestMetricExporterValidation` doesn't reset `envConfigCache`" — false alarm: `newMetricExporter` reads `os.Getenv` directly, not via the cache.
- "`captureStdout` is racy against parallel tests in the same package" — false alarm: Go's test runner serializes non-parallel tests before parallel ones, and no parallel test in the package writes to stdout.
- Several "function called but behavior not asserted" coverage nits — filtered as low-signal per heuristics.

---

*Generated by pr-review-curator skill. Filtered original ~12 findings down to 1 Major + 1 Info.*

internal/cli/command_integration_test.go
roark47: Good catch, this is valid. I updated `captureStdout` so the pipe and global `os.Stdout` are restored through a deferred cleanup path if `fn()` panics or exits early, while still waiting for the copy goroutine on the normal path. I also applied the related info-level cleanup suggestion for `serializeTranscript` by registering the returned cleanup with `t.Cleanup`.

Verified locally:
- `GOCACHE=/private/tmp/skill-up-gocache make fmt`
- `GOCACHE=/private/tmp/skill-up-gocache make verify`
- `GOCACHE=/private/tmp/skill-up-gocache make test`

zpzjzj (approved):
LGTM

PR #74, opened 2026-05-27
ci: refresh coverage badge on main events

roark47:
## Summary
- run CI on every push to main so coverage badge updates after docs-only or workflow-only merges too
- add workflow_dispatch so maintainers can manually refresh the coverage badge
- keep badge artifact upload and badges-branch write job restricted to main push/manual events, not PRs

## Verification
- ruby -e 'require "yaml"; YAML.load_file(".github/workflows/ci.yml")'
- git diff --check
- GOCACHE=/private/tmp/skill-up-gocache make verify
- GOCACHE=/private/tmp/skill-up-gocache make test


zpzjzj:
I rechecked this against current `main` before reviewing. The branch is now 80 commits behind, and the coverage-badge build/upload/commit jobs this patch modifies are no longer present in the current default `.github/workflows/ci.yml` after the CI split. As a result, the existing diff is no longer reviewable as an incremental change.

Please either close this PR if the original badge-refresh requirement is obsolete, or rework it from current `main` against the workflow that now owns coverage publishing. I’m holding off on a line-by-line review until that scope is refreshed.

PR #84, opened 2026-06-02
fix(runner): report skill name from SKILL.md frontmatter, not dir basename

zpzjzj:
## Problem

Evaluation reports (`result.json`, `benchmark.json`/`.md`, JUnit class name, HTML title) derived the skill name from the **skill directory's basename** via `filepath.Base(...)` in [`internal/runner/runner.go`](internal/runner/runner.go). The authoritative `name:` field in `SKILL.md`'s YAML frontmatter was never parsed anywhere in the codebase (`FindSkillDir` only `os.Stat`s the file for existence). So a skill whose directory name differs from its declared `name` showed the wrong identity in every report.

## Fix

- Add `Loader.SkillName()` in [`internal/config/loader.go`](internal/config/loader.go), which extracts the `name` field from `SKILL.md`'s YAML frontmatter. It returns `("", false)` when `SKILL.md` is missing, has no frontmatter fence, fails to parse, or declares an empty name.
- The runner derives a separate `reportName` that prefers the frontmatter name and **falls back to the directory basename** otherwise, then uses it for the report-facing consumers (`newEvaluator` and `WriteResults`).

## No impact on existing functional links

The dir-based name is **retained** for the workspace layout (`<skill>-workspace`) and the install paths — only the *reported* identity changed. When `SKILL.md` has no `name` (or is absent), behavior is identical to before.

## Tests

- `TestLoader_SkillName` (table-driven): frontmatter name differs from dir / no name field / empty name / no frontmatter fence / missing `SKILL.md`.
- Manually verified end-to-end: a skill in dir `tmp-xyz` with `name: code-stats` reports `code-stats` while the workspace dir stays `tmp-xyz-workspace`.
- `go build ./...`, `go vet`, `gofmt`, and `go test ./internal/config/... ./internal/runner/...` all pass.

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -2,9 +2,11 @@
 package config
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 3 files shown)


roark47 (approved):
LGTM

lijunfeng722 (approved):
LGTM

Issue #50, opened 2026-05-21
Custom Engine: extract template-resolution into a leaf customengine package


zpzjzj:
## Background

Two near-identical token parsers exist for the `\${X}` / `\${X:-d}` / `\${X?msg}` syntax:

- `internal/agent/custom.go::resolveTemplateToken` — run-time, against the rendered variable map.
- `internal/config/customengine.go::resolveEnvToken` — config-load-time, strict mode.

They differ only in strictness and return signature. Bugs fixed in one are easy to miss in the other.

Independently, `internal/config/CustomEngineConfig` is now imported by both the `config` and `agent` packages. The agentkind leaf package (#PR24) already broke the factory dispatch cycle, but the config type itself still lives in `internal/config`, which means anything that needs the type today must depend on `config`.

## Proposal

1. Introduce `internal/customengine` as a new leaf package that owns:
   - `CustomEngineConfig`, `CustomLocalConfig`, `CustomHTTPConfig`, `CustomHTTPFile` types (moved out of `internal/config/schema.go`).
   - A shared `ResolveToken(inner string, vars map[string]string, opts ResolveOpts)` so both call sites use one parser.
2. Re-export shims from `internal/config` for one release to avoid a hard break for any external importer.

## Why this is a follow-up, not a blocker

Both parsers behave correctly today and are covered by tests. The duplication is a maintenance hazard, not a defect.

## Origin

Raised during the self code-review of `feat/custom-engine-local` (code-quality finding #1 and architecture finding #2).

PR #89, opened 2026-06-08
refactor(custom-engine): share ${...} token parsing between config and agent resolvers (#50)

zpzjzj:
## Summary

Part 1 of #50. The config-load-time resolver (`config.resolveEnvToken`) and the agent-runtime resolver (`agent.resolveTemplateToken`) each hand-rolled the **same** `${X}` / `${X:-default}` / `${X?msg}` grammar split, so a fix to one was easy to miss in the other. This extracts the shared parse into `config.ParseTemplateToken` (+ `TemplateToken.RequiredErr` for the `${X?msg}` blank-message default) and has both resolvers call it — the grammar now has a single source of truth.

## Scope (important)

Investigation showed the issue's premise ("two near-identical parsers") only holds for the **grammar split**. The two resolvers otherwise differ substantially:

| | agent `resolveTemplateToken` | config `resolveEnvToken` |
|---|---|---|
| lookup | vars map + env fallback | env only |
| builtin vars | resolved as normal | left intact (`leaveIntact`) |
| empty value | present-but-empty → triggers default | n/a |
| `kwargs.*` missing | → empty string | n/a |
| secrets | none | multi-layer rejection |
| passes | single | strict: up to 10 |
| return | `(string, error)` | `(value, leaveIntact, error)` |

Forcing these into one `ResolveToken` (lookup callback + options + merged return) would be *more* complex, not less. So **only the genuinely-shared grammar is unified**; each resolver keeps its own lookup/secret/empty policy verbatim.

**Part 2** of #50 (extracting the four custom-engine config types into an `internal/customengine` leaf package) is deferred as a separate follow-up before HTTP transport — its current ROI is low (config still needs the types for `EngineConfig.Custom`). Hence `Refs #50`, not `Closes`.

## Where the helper lives

`internal/config` for now — it's already imported by `internal/agent` (no import cycle; config does not import agent). When part 2 lands, the helper moves to the `internal/customengine` leaf package alongside the types.

## Test plan

- [x] `make test` (`go test -race ./...`) — green. Existing parser tests (`TestRenderTemplate*` in agent, config env-ref resolution) are **unchanged** and pass (the behavior safety net).
- [x] New `internal/config/template_token_test.go` — covers the three forms, `:-` before `?` precedence, and the `RequiredErr` blank-message default.
- [x] `make verify` (fmt + vet + revive + golangci-lint) — **0 issues**.
- [x] `make e2e` custom-engine suite — `RejectsMissingEnvVar` (config strict path), `LocalTransport` (agent render path), `Validate` all pass.

## Why not an existing envsubst library?

Libraries exist for `${VAR}` / `${VAR:-default}` substitution (`os.Expand` in stdlib, `drone/envsubst`, `a8m/envsubst`, or the full `mvdan.cc/sh`), but they only cover the **grammar split** — which is ~6 lines. The parts that actually matter here are bespoke and can't be delegated to a generic substituter:

- **Custom lookup sources** — the agent resolver resolves against a runtime *vars map* (`case_id`, `kwargs.<key>`, `model`, …) with env only as fallback, not just `os.Getenv`.
- **Two-stage `leaveIntact` resolution** — the config resolver deliberately leaves builtin template vars (`${api_key}`, `${kwargs}`, …) untouched at load time for the agent to resolve at run time. No generic lib models this.
- **Secret-leak prevention** — strict mode rejects credentials reaching a command line via three layers (sensitive var name / credential-shaped `:-` default literal / credential-shaped resolved value) plus a multi-pass (≤10) anti-smuggling expansion so a non-sensitive wrapper var can't smuggle a sensitive name through. This security policy is the whole point and no envsubst lib provides it.

Pulling a third-party dependency to save ~6 lines — while that dependency *cannot* carry the credential-leak-prevention responsibility — is a poor trade, especially in a supply-chain-sensitive repo. So #50 keeps the resolvers in-house and only dedups the shared grammar split into a unit-tested `ParseTemplateToken`. If the grammar ever needs to converge on full POSIX (`:+`, `:=`, …), swapping in `a8m/envsubst` for the split would be worth revisiting then.

### DIFF
--- a/internal/agent/custom.go
+++ b/internal/agent/custom.go
@@ -1147,37 +1147,29 @@ func renderTemplate(s string, vars map[string]string) (string, error) {
 }
 
 func resolveTemplateToken(inner string, vars map[string]string) (string, error) {
-	name := inner
-	var defaultVal, errMsg string
-	hasDefault, hasErrForm := false, false
-	if before, after, found := strings.Cut(inner, ":-"); found {
-		name, defaultVal, hasDefault = before, after, true
-	} else if before, after, found := strings.Cut(inner, "?"); found {
-		name, errMsg, hasErrForm = before, after, true
-	}
+	// Share the ${X} / ${X:-default} / ${X?msg} grammar with the config-load-time
+	// resolver (config.resolveEnvToken) so the two cannot drift apart.
+	tok := config.ParseTemplateToken(inner)
 
-	if v, ok := vars[name]; ok {
+	if v, ok := vars[tok.Name]; ok {
 		// A present-but-empty built-in value (e.g. an unconfigured api_key)
 		// is treated as unset so ${X:-default} and ${X?msg} still apply.
-		if v != "" || (!hasDefault && !hasErrForm) {
+		if v != "" || (!tok.HasDefault && !tok.HasErrForm) {
 			return v, nil
 		}
 	}
-	if v := os.Getenv(name); v != "" {
+	if v := os.Getenv(tok.Name); v != "" {
 		return v, nil
 	}
-	if hasDefault {
-		return defaultVal, nil
+	if tok.HasDefault {
+		return tok.Default, nil
 	}
-	if hasErrForm {
-		if strings.TrimSpace(errMsg) == "" {
-			errMsg = name + " is required"
-		}
-		return "", errors.New(errMsg)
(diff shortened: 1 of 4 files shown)


zpzjzj:
@codex review

roark47 (approved):
LGTM

PR #93, opened 2026-06-08
test: assert custom engine report content in e2e tests

### LINKED ISSUE #53: Custom Engine e2e: assert generated report content, not just summary substring
## Background

The Custom Engine e2e tests (`e2e/custom_engine_test.go`) currently assert only that the CLI produced output containing the substrings `PASS` and `1 passed`. They do not verify:

- The custom engine's `final_message` actually reached the judge.
- `expect.must_contain` and other gate assertions evaluated correctly.
- The generated `report.json` payload matches what was actually executed.

A custom engine that returns the wrong `final_message` but happens to print `PASS` somewhere would pass these tests.

## Proposal

After the run, parse the generated `report.json` and assert:

- The case's `final_message` matches the fixture script's output.
- The case's `judge.outcome` is the expected pass/fail.
- The transcript role/content matches the user message that was fed in.

## Why this is a follow-up, not a blocker

The current e2e tests do catch the most important regression — "Custom Engine doesn't run at all" — but the safety net is shallow.

## Origin

Raised during the self code-review of `feat/custom-engine-local` (CI/tests finding #3).

zpzjzj:
## Summary
- Enhance `TestPipeline_CustomEngine_LocalTransport` to parse generated `report.json` after pipeline run
- Assert `final_message` matches fixture script output (`custom-engine-handled`)
- Assert judge outcome is `StatusPass`
- Assert input prompt contains the expected user message

Closes #53

## Test plan
- [ ] `go build ./...` passes
- [ ] `go vet ./...` passes
- [ ] E2e tests pass with `go test ./e2e/... -run TestPipeline_CustomEngine`

### DIFF
--- a/e2e/custom_engine_test.go
+++ b/e2e/custom_engine_test.go
@@ -3,12 +3,16 @@
 package e2e
 
 import (
+	"encoding/json"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/alibaba/skill-up/internal/judge"
+	"github.com/alibaba/skill-up/internal/report"
 )
 
 // getCustomEngineTestdataDir returns the path to e2e/testdata/custom-engine.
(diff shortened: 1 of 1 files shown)


roark47 (approved):
LGTM

Issue #51, opened 2026-05-21
Custom Engine: define internal transport interface seam before HTTP lands


zpzjzj:
## Background

`engine.custom.transport: http` is fully declared in the schema (`internal/config/schema.go` `CustomHTTPConfig`) and rejected at runtime with a clear "not yet implemented" error (`internal/agent/custom.go::Run`). The local-transport implementation, however, is currently a pile of `runLocal` / `buildLocalExec` / `finishLocal` / `readRawResult` private methods on `CustomAgent`, with assumptions baked in (output file on disk, stdout fallback, artifact paths in workspace).

When HTTP lands, those assumptions don't translate:

- Result format negotiation: HTTP body vs file read.
- Artifact handling: HTTP file downloads with retry vs local filesystem reads.
- Streaming responses.
- Retry / circuit-breaker policies.

## Proposal

Introduce an internal `transport` interface inside the agent package:

```go
type customTransport interface {
    Run(ctx context.Context, opts ExecOptions, vars map[string]string, sess *SessionInput) (rawResult string, execErr error, artifactSources []artifactSource)
}
```

Move the existing `runLocal` body behind a `localTransport` struct that satisfies this. `CustomAgent.Run` dispatches to the right transport without further branching.

Outcome: HTTP gains a parallel `httpTransport` that drops in without rewriting `finishLocal`.

## Why this is a follow-up, not a blocker

Doing it now would be premature — we don't yet know HTTP's real shape. Doing it just before HTTP implementation begins is the sweet spot.

## Origin

Raised during the self code-review of `feat/custom-engine-local` (architecture finding #3).

PR #98, opened 2026-06-08
refactor(runtime): move case metadata into AgentMetadata off generic ExecOptions (#49)

### LINKED ISSUE #49: Custom Engine: extract agent-specific metadata out of generic ExecOptions
## Background

When the Custom Engine landed (#24), three new fields were added directly to the generic `internal/runtime/runtime.go::ExecOptions`:

- `CaseID`
- `Variant`
- `MaxTurns`

with a comment noting "Built-in agents ignore these fields." This is implicit coupling — `ExecOptions` is the universal runtime parameter struct, but it now carries metadata that only `CustomAgent` consumes. The evaluator's `runAgent.Run(...)` call site populates these fields unconditionally (`internal/evaluator/evaluator.go:394`), and built-in agents silently discard them.

## Proposal

Bundle the case-level agent metadata into its own struct so the contract is explicit:

```go
// in internal/runtime
type AgentMetadata struct {
    CaseID   string
    Variant  string
    MaxTurns int
}

type ExecOptions struct {
    // ... existing fields ...
    AgentMetadata *AgentMetadata // nil for runs that don't need it
}
```

Alternatively, push the metadata through `agent.Config` at agent construction (once per case) rather than on every `Exec` call.

## Why this is a follow-up, not a blocker

The current shape works and is observable; no built-in agent misuses the fields. But the next agent type that needs different case-level metadata will compound the smell. Address it before HTTP transport or a second user-defined-agent kind lands.

## Origin

Raised during the self code-review of `feat/custom-engine-local` (architecture dimension, finding #1).

zpzjzj:
## Summary

Closes #49. `CaseID` / `Variant` / `MaxTurns` were added directly to the universal `runtime.ExecOptions` when the Custom Engine landed (#24), with a comment that "built-in agents ignore these fields" — implicit coupling: a generic exec struct carrying metadata only `CustomAgent` consumes. This bundles them into a dedicated `runtime.AgentMetadata` threaded via `ExecOptions.AgentMetadata` (a pointer, nil when unused), so the contract is explicit.

## Why Option A (pointer on ExecOptions), not Option B (agent.Config at construction)

The issue offered two designs. Option B doesn't fit: the evaluator holds a **single agent instance and reuses it across cases** (`runAgent := e.ag` in `executeCaseOnce`), so per-case metadata can't be baked into the agent at construction — it must travel per-`Run`. `ExecOptions` is the per-Run parameter, so the metadata belongs there.

## Changes

- **runtime** (`runtime.go`): add `AgentMetadata{CaseID,Variant,MaxTurns}` + a nil-safe `ExecOptions.AgentMeta()` accessor; drop the three bare fields from `ExecOptions`.
- **evaluator** (`evaluator.go`): the sole writer wraps the trio into `&runtime.AgentMetadata{...}`.
- **agent** (`custom.go`): the six read sites (`buildSessionInput`, `buildBaseVars`) read via `opts.AgentMeta()`.

## Behavior-preserving

`AgentMeta()` returns a zero value when `AgentMetadata` is nil — matching the old "fields unset" semantics (empty `CaseID`/`Variant`, `MaxTurns` 0). Built-in agents never read these fields, so nothing changes for them.

This is a struct-shape refactor, so unlike the previous pure-internal refactors it touches a few test constructions (`custom_test.go`) that set the fields — `evaluator_test.go` was confirmed unaffected (its `CaseID` is on a different struct).

## Test plan

- [x] `make test` (`go test -race ./...`) — runtime, agent, evaluator packages green.
- [x] `make verify` (fmt + vet + revive + golangci-lint) — **0 issues**.
- [x] `make e2e` custom-engine suite — `LocalTransport` / `Validate` / `RejectsMissingEnvVar` all pass (exercises the CaseID/Variant/MaxTurns flow end-to-end).

## Context

This is the last small prerequisite flagged "before HTTP transport lands" (#51 transport seam ✅ merged; #50 part 1 parser dedup ✅ merged). The HTTP transport will consume the same per-case metadata, so tightening the contract now avoids the smell compounding across a second transport.

### DIFF
--- a/internal/agent/custom.go
+++ b/internal/agent/custom.go
@@ -894,14 +894,15 @@ func (a *CustomAgent) buildSessionInput(rt Runtime, opts ExecOptions, messages [
 	for _, m := range messages {
 		msgs = append(msgs, SessionMessage{Role: string(m.Role), Content: m.Content})
 	}
+	meta := opts.AgentMeta()
 	return SessionInput{
-		CaseID:         opts.CaseID,
-		Variant:        opts.Variant,
+		CaseID:         meta.CaseID,
+		Variant:        meta.Variant,
 		Workspace:      rt.Workspace(),
 		Model:          formatAgentModel(a.Cfg.ModelProvider, a.Cfg.ModelName),
 		Kwargs:         kwargs,
 		Messages:       msgs,
-		MaxTurns:       opts.MaxTurns,
+		MaxTurns:       meta.MaxTurns,
 		TimeoutSeconds: timeoutSec,
 	}
 }
@@ -912,6 +913,7 @@ func (a *CustomAgent) buildSessionInput(rt Runtime, opts ExecOptions, messages [
 // once resolveCustomIOFiles has run.
 func (a *CustomAgent) buildBaseVars(rt Runtime, opts ExecOptions, messages []transcript.Message, timeoutSec int) map[string]string {
 	workspace := rt.Workspace()
+	meta := opts.AgentMeta()
 	return map[string]string{
 		"workspace":       workspace,
 		"prompt":          singleTurnPrompt(messages),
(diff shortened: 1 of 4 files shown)


zpzjzj:
@codex review

roark47 (approved):
LGTM

PR #112, opened 2026-06-22
docs: correct the action's uses: ref (v0.1.0 was a nonexistent tag)

zpzjzj:
## What

The GitHub Action usage examples in `README.md` / `README.zh.md` referenced `uses: alibaba/skill-up@v0.1.0`. That is wrong:

- `v0.1.0` is the runner **image** tag (`ghcr.io/alibaba/skill-up-runner:v0.1.0`), **not a git tag** — no such release/tag exists.
- Release tags published **before** the action was merged (≤ `v0.2.4`) don't contain `action.yml`, so they can't be used as the ref either.

A consumer copy-pasting the example would hit a "can't find action.yml" error.

## Change

- Point the examples at `@main` (works today).
- Add a **Versioning** note: pin a release tag that includes the action (or a commit SHA) for stability; `@main` tracks latest; pre-action tags don't work.
- Add the caller requirements (Linux runner + a model-key secret; the runner image is public so no extra registry auth).

Docs-only; `action.yml`'s image reference (`:v0.1.0`) is unchanged.

> Note: the action is fully usable via `uses: alibaba/skill-up@<ref>` regardless of a Marketplace listing — Marketplace is discoverability only, and publishing it is currently gated on the org accepting the Marketplace Developer Agreement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,17 +314,29 @@ jobs:
     runs-on: ubuntu-latest          # Docker container action — Linux only
     steps:
       - uses: actions/checkout@v4
-      - uses: alibaba/skill-up@v0.1.0
+      - uses: alibaba/skill-up@main  # see "Versioning" below
         with:
           engine: claude_code        # or codex / qodercli; empty = let eval.yaml decide
           api-key: ${{ secrets.ANTHROPIC_API_KEY }}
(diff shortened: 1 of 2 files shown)


zpzjzj:
@codex review

roark47 (approved):
LGTM

PR #127, opened 2026-07-03
fix: preserve exec perms on skill install; validate duplicate case IDs (#118, #123)

### LINKED ISSUE #118: Preserve executable file permissions when installing skills
## Background

During the `app-upgrade-agents` migration to skill-up, shell scripts inside the skill lost their executable bit after skill installation. This made agents fail immediately with missing/non-executable script errors, and forced skill-specific permission repair workarounds.

This is a general skill-up problem for any skill that ships executable helpers, not specific to app-upgrade-agents.

## Why skill-up should change

Skill installation is owned by skill-up, so installed files should preserve the executable semantics of the source skill. Today, skill installation uploads files one-by-one through the runtime transfer API, and the local `none` runtime writes uploaded files with a fixed file mode. That can turn an executable source script into a non-executable installed script.

## What to change

- Preserve source file mode when uploading/installing skill files, especially executable bits on scripts.
- Make the runtime file-transfer contract explicit about permission preservation.
- Ensure behavior is consistent across `none` and `opensandbox` runtimes where the target runtime supports file modes.
- Add focused tests for skill installation, not only generic directory copy behavior.

## How to verify

- Create a fixture skill with `scripts/run.sh` set to `0755`.
- Install it into a `none` runtime and assert the installed file remains executable.
- Run the installed script through the target agent/runtime and assert it succeeds without any chmod workaround.
- Add or update unit tests around runtime upload and agent skill installation.

## Architecture impact

This affects the runtime transfer contract and skill installation behavior. The fix should avoid baking shell-specific logic into agent adapters; permission preservation belongs in runtime/file transfer or skill install plumbing.

### LINKED ISSUE #123: Validate duplicate case IDs and duplicate case file references
## Background

The `app-upgrade-agents` migration hit configuration conflicts caused by repeated case definitions. Duplicate cases are easy to create while splitting smoke/full suites, copying YAML files, or maintaining large eval sets.

This is a general data integrity issue because skill-up writes reports and artifacts keyed by case ID.

## Why skill-up should change

Case IDs should be unique within an eval. If two case files resolve to the same ID, reports and artifacts can be overwritten, merged, or become ambiguous. Duplicate entries in `cases.files` can also cause the same case to run more than once unintentionally.

The validator should catch this before execution starts.

## What to change

- Add validation that every loaded case has a unique effective `id`.
- Add validation that `cases.files` does not list the same file more than once after path normalization.
- Include the conflicting case ID and source file paths in the error message.
- Keep filename-derived case IDs covered, not only explicit `id:` fields.

## How to verify

- Create two case YAML files with the same explicit `id`; `skill-up validate` should fail with a clear duplicate ID error.
- Create two files whose filename-derived IDs collide, or repeat the same file in `cases.files`; validation should fail before execution.
- Confirm a valid suite with unique IDs still runs normally.
- Add validator and loader tests covering explicit IDs, implicit filename IDs, and duplicate file references.

## Architecture impact

This is a low-risk config validation improvement. It protects runner/report assumptions that group and write artifacts by case ID.

zpzjzj:
Follow-up fixes surfaced by the `app-upgrade-agents` migration. Two independent, low-risk changes bundled into one PR (one commit each).

## #118 — Preserve executable file permissions when installing skills

`Closes #118`

The `none` runtime's per-file `UploadFile` wrote every file with a fixed `0600` mode, so a skill shipping runnable helper scripts (e.g. `scripts/run.sh` at `0755`) installed **non-executable** and failed at runtime without a chmod workaround.

- `UploadFile` now stats the source and writes with its permission bits, then `Chmod`s to make the executable bit deterministic under any umask or on re-install.
- The `Runtime` file-transfer contract now documents permission preservation.
- No behavior change for other callers: `fixtures`/`agent` upload temp files created by `os.CreateTemp` (already `0600`); `opensandbox` (stats each file) and `docker` (`docker cp`) already preserved mode.

## #123 — Validate duplicate case IDs and duplicate file references

`Closes #123`

Reports and artifacts are keyed by case ID and one case runs per `cases.files` entry, so a collision would silently overwrite results or double-run a case.

- `ValidateAll` now rejects duplicate effective case IDs (covering both explicit `id:` and filename-derived IDs) and duplicate `cases.files` entries (collapsing differently-spelled paths via `path.Clean`).
- Each error names the conflicting ID and the offending source files. `result.Cases` is index-aligned with `Eval.Cases.Files`, so no change to `CaseConfig` was needed.

## Testing

- `go test -race ./internal/runtime/ ./internal/config/ ./internal/agent/` — green.
- New tests: runtime + install-level executable-bit preservation; validator subtests for explicit-id / implicit-id / duplicate-file / unique-pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Skill installation preserves source file permissions, notably the executable
+  bit on scripts. The `none` runtime's per-file upload previously wrote every
+  file with a fixed `0600` mode, so skills shipping runnable helper scripts
+  installed non-executable and failed without a chmod workaround. The
+  file-transfer contract now documents permission preservation, matching the
+  `opensandbox` and `docker` runtimes which already carried the mode across.
+
 ## [0.3.0] - 2026-07-03
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Config validation now rejects duplicate case IDs and duplicate `cases.files`
+  references. Since reports and artifacts are keyed by case ID and one case runs
+  per `cases.files` entry, a collision would silently overwrite results or
+  double-run a case. `skill-up validate` (and the preflight in `skill-up run`)
+  fails with the conflicting ID and the offending source files. Covers both
+  explicit `id:` fields and filename-derived IDs, and collapses
+  differently-spelled paths (e.g. `cases/a.yaml` vs `./cases/a.yaml`).
+
 ### Fixed
 - Skill installation preserves source file permissions, notably the executable
   bit on scripts. The `none` runtime's per-file upload previously wrote every
--- a/internal/agent/agent_test.go
+++ b/internal/agent/agent_test.go
@@ -6,6 +6,7 @@ import (
 	"maps"
 	"os"
 	"path/filepath"
+	goruntime "runtime"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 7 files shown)


roark47 (approved):
LGTM

PR #131, opened 2026-07-06
feat(cli): validate only selected cases in run (#121)

### LINKED ISSUE #121: Allow run filters to validate only selected cases or support first-class eval profiles
## Background

During the `app-upgrade-agents` migration, smoke tests and full end-to-end tests had to be split into separate eval files. Otherwise, configuration intended only for the full suite could block running the smoke subset.

This hurts maintainability because smoke and full evaluations often share the same skill, environment, engine, and fixtures.

## Why skill-up should change

`skill-up run` supports case filters, but the current flow validates the loaded eval and all referenced cases before applying include/exclude filters. That means an unselected case can prevent a filtered run from starting.

For day-to-day CI, users need to run a quick subset without being blocked by unrelated cases that are not part of that run. At the same time, `skill-up validate` should continue to validate the whole suite.

## What to change

- Either validate only selected cases during `skill-up run` after include/exclude filters are applied, or add first-class profiles/tags such as `--include-tag smoke` and validate the selected profile.
- Keep `skill-up validate` as the command that performs full-suite validation.
- Make the behavior explicit in CLI docs.
- Ensure judge config merging and case-level judge overrides are validated against the effective selected cases.

## How to verify

- Create one eval containing both `smoke` and `full` cases.
- Make the full case intentionally invalid, for example with an incomplete script judge config.
- Run only the smoke case/profile and verify it executes successfully.
- Run `skill-up validate` and verify it still reports the invalid full case.
- Add CLI tests covering filter-before-run-validation behavior.

## Architecture impact

This changes the ordering of config loading, filtering, and validation for `run`. The distinction between `run` validation and `validate` validation should be documented so users know which command checks the full suite.

zpzjzj:
Fixes #121.

## Problem

`skill-up run` validated the **whole** suite (eval config + every case) *before* applying `--include-case-name` / `--exclude-case-name` filters. So one invalid unselected case blocked a filtered run — forcing teams to split smoke and full suites into separate eval files even when they share the same skill/environment/engine/fixtures.

## Change

- `run` now validates the **eval-level** config at load, then validates **only the cases left after filtering**. An invalid case that is filtered out no longer blocks the run.
- `skill-up validate` is unchanged and still validates the **whole** suite (every case) regardless of filters.

### Implementation

- Split `Validator.ValidateAll` into `ValidateEvalConfig` (suite-level) + new **`ValidateCases(cases)`** (duplicate-ID / duplicate-file + per-case checks over whatever set is passed). `ValidateAll = ValidateEvalConfig + ValidateCases(all)`, so `validate` behavior is identical.
- Add **`CaseConfig.SourceFile`** (loader-populated, `yaml:"-" json:"-"`). Duplicate-case detection now keys off `ID` + `SourceFile`, so it is correct for any subset — and this removes the fragile `result.Cases ↔ Eval.Cases.Files` index-alignment the #123 check relied on.
- `run.go`: eval-level validation in `loadFromEvalYAML`; `ValidateCases` over the filtered set in `loadAndPrepareConfig`.
- Docs: `run`-vs-`validate` validation scope documented in the command help and `cli-reference.md`.

Note: the `--auto` / `evals.json` run path now also validates its cases (previously unvalidated). This is safe — converted cases carry prompts, and their empty `SourceFile` skips the file-duplicate check.

## Verify (matches the issue's steps)

- New `internal/cli` test: an eval with a valid `smoke` case + an invalid `full` case (script judge, no `script_path`). `run --include-case-name smoke` succeeds; unfiltered `run` fails with the `script_path` error.
- New validator tests: `ValidateCases` scoped duplicate detection; `ValidateCases([valid])` passes while `ValidateCases([valid, invalid])` fails.
- `go test -race ./...` green; `gofmt` / `go vet` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- `skill-up run` now validates only the cases left after
+  `--include-case-name` / `--exclude-case-name` filters (plus the eval-level
+  config), so an invalid case that is filtered out no longer blocks a filtered
+  run — a shared eval can hold a quick `smoke` subset alongside heavier cases.
+  `skill-up validate` is unchanged and still validates the whole suite (every
+  case) regardless of filters.
+
 ### Added
 - Config validation now rejects duplicate case IDs and duplicate `cases.files`
   references. Since reports and artifacts are keyed by case ID and one case runs
--- a/docs/guide/cli-reference.md
+++ b/docs/guide/cli-reference.md
@@ -34,6 +34,13 @@ skill-up run [path] [flags]
 | `--api-key`            | —                             | Pass an API key (higher precedence than env vars)                                                                                                          |
 | `-v, --verbose`        | `0`                           | Increase log verbosity. Default `info`; `-v` / `--verbose` / `--verbose=true` → `debug`; `-vv` / `--verbose=2` → `trace`; `--verbose=false` disables extra detail |
 
+> **Validation scope.** `run` validates the eval-level config plus **only the
+> cases selected** after `--include-case-name` / `--exclude-case-name` filters.
+> An invalid case that is filtered out does not block the run — so a shared
+> eval can hold both a quick `smoke` subset and heavier cases without the latter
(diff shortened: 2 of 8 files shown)


roark47 (approved):
LGTM

PR #132, opened 2026-07-06
fix: gracefully terminate none-runtime process group on cancel

### LINKED ISSUE #120: Clean up process groups on none runtime timeout or cancellation
## Background

The `app-upgrade-agents` migration exposed a failure mode where long-running helper scripts and background processes could survive cancellation, leaving stale locks or zombie-like process state. In one case, the evaluation waited until a long timeout because a lock mechanism still believed a previous process was alive.

This is a general risk for end-to-end skill evaluations that run build tools, package managers, shell helpers, or background subprocesses.

## Why skill-up should change

The runtime owns command execution and timeout enforcement. When skill-up cancels or times out a command, the runtime should make a best effort to terminate the full process tree/process group, not just the immediate shell process.

## What to change

- For Unix-like `none` runtime execution, start commands in a separate process group.
- On context cancellation or timeout, terminate the process group, with an escalation path if graceful termination does not finish promptly.
- Preserve the current `ExecResult` behavior for stdout, stderr, and exit status as much as possible.
- Add platform-specific build tags or fallbacks where needed.

## How to verify

- Add a test command that starts a background child process and then blocks beyond a short context timeout.
- After timeout, assert the child process is no longer running.
- Add a lock-file style regression test: the first timed-out command starts a child that would keep a lock alive; a second command should not hang on that stale child.
- Run the runtime tests on supported Unix platforms.

## Architecture impact

This changes the `none` runtime execution semantics around cancellation. It should be implemented at the runtime layer so agent adapters and skill authors do not need to implement their own process-tree cleanup logic.

zpzjzj:
## Summary

The `none` runtime already starts commands in their own process group and kills the whole group on cancellation, but it did so with an **immediate SIGKILL**. Children never got a chance to run trap handlers, release locks, or clean up — the exact failure mode the issue describes, where a helper holding a lock left stale state that stalled a later command.

This adds a graceful escalation path:

- On cancel/timeout, the process group is signaled with **SIGTERM first**, giving children a chance to clean up.
- If the group is still alive after a short grace (`noneExecKillGrace = 1s`), it escalates to **SIGKILL**.
- The grace fits inside the existing 2s `WaitDelay`, so a timed-out `Exec` still returns within its deadline and preserves the `context.DeadlineExceeded` / exit-code `-1` contract.
- Whole-group semantics (negative PID) and the single-process fallback are kept. Escalation runs out of band (goroutine) so `cmd.Cancel` returns immediately and `os/exec` can start its `WaitDelay` timer; a `done` channel closed once `Wait` returns stops the escalation timer before the PID can be recycled.
- No-op on non-Unix platforms (`none_exec_other.go`, guarded by the same build tags).

## Tests

- `TestNoneRuntime_ExecTerminatesGracefullyThenEscalates` — a SIGTERM trap writes a marker before exit, proving the graceful path runs cleanup rather than an immediate SIGKILL (verified: fails under immediate-SIGKILL, passes with SIGTERM-first).
- `TestNoneRuntime_ExecStaleLockDoesNotHangNextCommand` — the issue's lock-file regression: a first command times out while a background child records its PID as a lock; after the timeout the child's group is gone, and a second command sees a dead PID (`kill -0`) and does not hang.
- Existing `TestNoneRuntime_ExecKillsDescendantsOnTimeout` and the deadline/cancel contract tests still pass.

`go build ./...`, `go vet ./...`, `gofmt -l` clean, and `go test -race ./...` green.

Fixes #120

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   explicit `id:` fields and filename-derived IDs, and collapses
   differently-spelled paths (e.g. `cases/a.yaml` vs `./cases/a.yaml`).
 
+### Changed
+- The `none` runtime now terminates a canceled or timed-out command's whole
+  process group gracefully before force-killing it. On Unix it sends `SIGTERM`
+  to the group first, giving children ~1s (`noneExecKillGrace`) to run trap
+  handlers, release locks, and exit, then escalates to `SIGKILL` if the group
+  is still alive. The grace fits inside the existing 2s `WaitDelay`, so a
+  timed-out `Exec` still returns within its deadline and keeps the
+  `context.DeadlineExceeded` / exit-code `-1` contract. Previously the group
+  got an immediate `SIGKILL`, so a helper holding a lock could leave stale
+  state and stall a later command. No-op on non-Unix platforms.
+
 ### Fixed
 - Skill installation preserves source file permissions, notably the executable
   bit on scripts. The `none` runtime's per-file upload previously wrote every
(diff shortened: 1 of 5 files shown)


roark47 (approved):
LGTM

PR #146, opened 2026-07-14
refactor(custom-engine): extract leaf config package

zpzjzj:
## Summary

- add a dependency-free `internal/customengine` leaf package for custom-engine config types, template-token grammar, and workspace-relative path checks
- make `internal/config` consume the canonical leaf types while retaining compatibility aliases and explicit wrappers for existing callers
- migrate custom agent transports to the leaf package, removing the `internal/agent -> internal/config` dependency without changing the YAML contract or runtime behavior

This completes the leaf-package extraction proposed in #50. The shared template-token parser landed earlier in #89; this PR moves that parser to its intended final home together with the remaining shared primitives.

## Verification

- `make verify`
- `make test`
- `go test -race ./internal/customengine ./internal/config ./internal/agent`
- `go test -tags e2e -count=1 -v -run '^TestPipeline_CustomEngine_(LocalTransport|HTTPTransport|Validate|RejectsMissingEnvVar)$' ./e2e`\n- import-graph assertion confirming `internal/agent` no longer imports `internal/config`\n\nCloses #50

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -11,8 +11,8 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/alibaba/skill-up/internal/config"
 	"github.com/alibaba/skill-up/internal/credential"
+	"github.com/alibaba/skill-up/internal/customengine"
 	"github.com/alibaba/skill-up/internal/logging"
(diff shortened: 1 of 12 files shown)


roark47 (approved):
LGTM

Issue #48, opened 2026-05-21
E2E coverage for agent CLIs on native Windows


zpzjzj:
**Context** (PR #33 self-review follow-up): the agent CLI path (Claude Code / Codex / Qoder) is launched via a bash-based Node/nvm bootstrap, so \`internal/agent\` on native Windows currently has zero e2e coverage. The Windows e2e job (\`E2E (none runtime, Windows)\`) only exercises mock-engine + script-judge.

**Proposal**:
- Either: rework the Node/nvm bootstrap to run under PowerShell, then enable the existing e2e suite on Windows.
- Or: add a Windows-WSL job that boots WSL, installs the agent CLIs, and runs the existing full e2e — covers the recommended workflow.
- Or (interim): add a Windows-host job that requires Git Bash pre-installed and runs full e2e through bash — covers the documented limitation path.

Currently the recommended workflow ("use WSL2") has no CI verification.

PR #148, opened 2026-07-17
feat(runtime): support Windows OpenSandbox guests

### LINKED ISSUE #45: OpenSandbox Windows guest profile (wait on Go SDK Platform field)
Once OpenSandbox-team/OpenSandbox#921 merges (Go SDK gains `PlatformSpec{OS, Arch}` on `CreateSandboxRequest` and `SandboxInfo`), wire up Windows guest sandboxes:

- `runtime.Config` accepts a `Platform` field (already documented as the future shape in `docs/guide/windows.md`).
- `OpenSandboxRuntime` passes `platform.os = windows` / `arch = amd64` through to the OpenSandbox API.
- `OpenSandboxRuntime.TargetGOOS()` (or its replacement once #44 lands) returns the requested guest OS instead of the current hardcoded `"linux"`.
- Add an opt-in e2e job (`E2E (opensandbox runtime, Windows guest)`) gated on the OpenSandbox endpoint actually serving Windows guests.

**Blocked on**:
- Upstream Go SDK PR: OpenSandbox-team/OpenSandbox#921 (typed `PlatformOS` / `PlatformArch` enums + `PlatformSpec`). Filed during PR #33; pending upstream review.

**Coordinates with #44**: once `Runtime.Shell()` lands, this issue's "return non-linux GOOS" wiring should happen there, not on the standalone `TargetGOOS()` method.

**Tracked from**: PR #33 self-review (deferred from the original scope of #31, which proposed both `none` + `opensandbox` Windows support).

zpzjzj:
## Summary

- add typed OpenSandbox `platform` and optional `resources` configuration with Windows guest defaults and user overrides
- route workspace setup, transfers, Custom Engine execution, PowerShell script judging, and artifact collection through guest-OS path and shell semantics
- replace the lifecycle compatibility implementation with the official Go SDK create path
- add bilingual documentation plus an opt-in Extended CI job that runs a deterministic real Windows guest lifecycle

## Compatibility

Omitting `environment.platform` and `environment.resources` preserves the existing Linux behavior. Built-in Agent CLI bootstrap inside Windows guests remains out of scope (tracked by #48); preinstalled CLIs and Custom Engines are supported.

## Validation

- `make verify`
- `go test -race ./internal/config ./internal/agent ./internal/judge ./internal/evaluator`
- `go test -race ./internal/runtime -skip TestNoneRuntime_ExecTerminatesGracefullyThenEscalates`
- `go test -tags e2e ./e2e -run "TestCustomEngine_OpenSandboxWindowsGuest|TestPipeline_CustomEngine_LocalTransport" -count=1`
- Windows fixture validation runs without infrastructure; the real guest run is gated by the Extended CI `windows_guest` workflow input and requires KVM/TUN/Docker

`make test` also exercised the full suite, but the existing 300 ms SIGTERM timing test `TestNoneRuntime_ExecTerminatesGracefullyThenEscalates` intermittently missed its trap under the race run. A subsequent runtime race run excluding only that known flaky passed. The local full quick E2E run additionally hit a broken machine-local Codex install (`spawn .../codex ENOENT`); the targeted Custom Engine E2E passed.

Closes #45


zpzjzj:
Extended CI follow-up: https://github.com/alibaba/skill-up/actions/runs/29551950851

- Linux OpenSandbox E2E passed after isolating the Windows server on port 8081.
- none-runtime full E2E, Windows-host quick E2E, Docker E2E/full-LLM, and GoReleaser checks passed.
- The real OpenSandbox Windows guest lifecycle was **not executed**: the selected runner reported `/dev/kvm` as unavailable or not readable/writable, so the gated steps were skipped. A KVM-capable runner is still required for end-to-end Windows guest validation.

roark47:
@copilot resolve the merge conflicts in this pull request

.github/workflows/extended-ci.yml
```diff
@@ -48,6 +54,109 @@ jobs:
           if-no-files-found: ignore
           retention-days: 14
 
+  e2e-opensandbox-windows:
+    name: E2E (OpenSandbox Windows guest)
+    if: github.event_name == 'workflow_dispatch' && inputs.windows_guest
+    runs-on: self-hosted
```
roark47, line 60: [P1] Please scope this to a dedicated trusted KVM-capable runner instead of the generic `self-hosted` pool. This job checks out and executes the selected ref, installs packages, starts Docker with host networking, and accesses `/dev/kvm` and `/dev/net/tun`; scheduling it on any self-hosted runner unnecessarily broadens the blast radius. Labels such as `[self-hosted, linux, x64, docker, trusted, kvm]` would also prevent the core test from silently landing on an incapable runner.

.github/workflows/extended-ci.yml
```diff
@@ -48,6 +54,109 @@ jobs:
...
+              echo "::error::OpenSandbox server did not become healthy within 120s"
+              cat "$log"
+              exit 1
+            fi
+            sleep 2
+          done
+          {
+            echo "OPENSANDBOX_API_KEY=$api_key"
+            echo "OPENSANDBOX_BASE_URL=http://127.0.0.1:8081"
+            echo "OPENSANDBOX_WINDOWS_IMAGE=dockurr/windows:latest"
```
roark47, line 132: [P1] Please pin the Windows image by digest rather than using `dockurr/windows:latest`. This image runs on a privileged, KVM/TUN-capable self-hosted runner with host networking, so a mutable tag creates a meaningful supply-chain and reproducibility risk.

internal/runtime/opensandbox.go
```diff
@@ -690,26 +707,43 @@ func (r *OpenSandboxRuntime) ensureDirectory(ctx context.Context, dir string, mo
...
+	command = "if not exist " + quotedNUL + " mkdir " + quotedDir +
+		" & if not exist " + quotedNUL + " exit /b 1" +
+		" & >" + quotedProbe + " echo probe" +
+		" & if not exist " + quotedProbe + " exit /b 1" +
+		" & del /q " + quotedProbe
+	return command, windowsVolumeRoot(dir)
+}
+
+func quoteCmdPath(value string) string {
+	return `"` + strings.ReplaceAll(value, "%", "%%") + `"`
```
roark47, line 739: [P1] This is not sufficient cmd.exe quoting for a configuration-controlled path. Doubling `%` does not handle embedded `"`, `&`, `|`, `<`, `>`, `^`, or delayed-expansion `!`; an input such as a crafted `workspace_mount` can terminate the quoted argument and alter the fallback command when `CreateDirectory` fails. Please use the shared cmd quoter if it covers path operands, or reject/escape cmd metacharacters here, and add adversarial tests.

internal/config/validator.go
```diff
@@ -560,6 +561,84 @@ func isValidRuntimeType(t string) bool {
...
+	if osName != "" && osName != "linux" && osName != "windows" {
+		errs = append(errs, "environment.platform.os must be one of: linux, windows")
+	}
+	if arch != "" && arch != "amd64" && arch != "arm64" {
+		errs = append(errs, "environment.platform.arch must be one of: amd64, arm64")
+	}
+	return errs
+}
+
+func validateOpenSandboxResources(env Environment) []string {
```
roark47, line 614: [P2] `cpu`, `memory`, and `disk` are only checked for whitespace-only values, so arbitrary strings and unbounded requests pass validation and fail later—or consume excessive resources if accepted by the server. Please validate the SDK/server-supported syntax and document/enforce where quotas live, or explicitly document that these opaque values are intentionally delegated to a quota-enforcing OpenSandbox server.

roark47 (changes requested):
Thanks for the comprehensive implementation. I checked the current head locally: `make verify` and the full `make test` race suite both pass.

I don't think this is ready to merge yet:

- The real OpenSandbox Windows guest lifecycle has not completed successfully. The documented Extended CI run skipped it because the selected runner lacked usable KVM. Since Windows guest support is the core behavior of this PR, please obtain at least one successful end-to-end run on a trusted KVM/TUN-capable runner.
- Please address the self-hosted runner scoping and mutable Windows image findings below.
- Please harden the cmd.exe path fallback quoting described below.
- Resource values are currently validated only for non-blank text. Please either validate the accepted CPU/memory/disk syntax (and document that quotas are enforced server-side), or explicitly justify delegating all validation to OpenSandbox.
- Before merging, the PR also needs to leave Draft state, be updated with main, obtain review approval, and clear the currently pending CLA check caused by the Copilot conflict-resolution commits.

The overall architecture—guest-specific path handling, traversal checks, SDK create path, and cross-component Windows adaptations—looks solid; the request for changes is about closing the security and validation gaps before merging.

pfangueiro (approved):

PR #151, opened 2026-07-23
feat(cli): add baseline run flag

googs1025:
## Summary
- Add `skill-up run --baseline` to temporarily enable benchmark mode from the CLI.
- Map the flag to `evalCfg.Benchmark.Enabled = true` while preserving config defaults when unset.
- Document the new flag in English/Chinese docs and skill-upper references.

## Test Plan
- `GOPROXY=https://goproxy.cn,direct make verify`
- `GIT_CONFIG_GLOBAL=/dev/null make test`


roark47 (approved):
Thanks for the contribution! The implementation is clean, focused, and well covered by tests and documentation. LGTM!

PR #170, opened 2026-08-04
fix(qoder): match underscore workspace keys

chiga0:
## Summary

- normalize underscores in Qoder project workspace keys
- keep the transform agent-specific so other session lookup behavior is unchanged
- add coverage for slash and underscore path encoding

## Validation

- `make fmt`
- `make verify`
- `env -u QODER_PERSONAL_ACCESS_TOKEN make test`
- real Qoder evaluation now resolves the native JSONL transcript; DataAgent base eval: 12/12 PASS, real integration eval: 16/16 PASS

### DIFF
--- a/internal/agent/session_lookup.go
+++ b/internal/agent/session_lookup.go
@@ -83,6 +83,9 @@ type agentSessionLookup struct {
 	// findExtra appends extra `find` predicates such as exclusion patterns.
 	// Empty string means no extra predicates.
 	findExtra string
+	// workspaceKeyTransform applies agent-specific path encoding after the
+	// shared runtime workspace key has been normalized.
+	workspaceKeyTransform func(string) string
 }
 
 // findAgentSessionJSONL resolves the newest *.jsonl session file under the
@@ -96,6 +99,9 @@ func findAgentSessionJSONL(ctx context.Context, rt Runtime, lookup agentSessionL
 	workspaceKey := workspaceKeyForRuntime(rt)
 	if workspaceKey == "" {
 		return ""
+	}
+	if lookup.workspaceKeyTransform != nil {
+		workspaceKey = lookup.workspaceKeyTransform(workspaceKey)
 	}
 	logging.DebugContextf(ctx, "agent session lookup: workspace=%q key=%q", rt.Workspace(), workspaceKey)
 
--- a/internal/agent/qodercli.go
+++ b/internal/agent/qodercli.go
@@ -280,10 +280,17 @@ func (a *QoderCLIAgent) RunTurn(ctx context.Context, rt Runtime, opts ExecOption
 // are read only inside the runtime via Exec (not os.Getenv / host os.Stat).
 func findQoderSessionFile(ctx context.Context, rt Runtime) string {
 	return findAgentSessionJSONL(ctx, rt, agentSessionLookup{
-		envVar:    "SKILL_UP_QODER_WSKEY",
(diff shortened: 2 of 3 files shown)


internal/agent/session_lookup.go
```diff
@@ -83,6 +83,9 @@ type agentSessionLookup struct {
 	// findExtra appends extra `find` predicates such as exclusion patterns.
 	// Empty string means no extra predicates.
 	findExtra string
+	// workspaceKeyTransform applies agent-specific path encoding after the
+	// shared runtime workspace key has been normalized.
+	workspaceKeyTransform func(string) string
```
zpzjzj, line 88: **🟠 Major: Make workspace-key encoding fully agent-owned**

`workspaceKeyTransform func(string) string` only post-processes `workspaceKeyForRuntime`, but that shared result already embeds Claude-specific path semantics. As a result, the new abstraction is still incomplete:

- Qoder replaces more than underscores. Existing Qoder project directories encode punctuation such as `.codex` with `-`, while this patch preserves the dot on POSIX.
- Qwen uses the same shared lookup without a transform, although upstream `sanitizeCwd` replaces every non-alphanumeric character with `-` and lowercases Windows paths. An underscore-containing Qwen workspace therefore has the same silent session-lookup/token-accounting failure class.

Please make the lookup accept a complete agent-owned encoder, e.g. `type workspaceKeyEncoder func(Runtime) string`, and require Claude, Qoder, and Qwen to each provide their full key implementation. The shared layer should continue to own session discovery, Windows `cygpath` conversion, download, and parsing; shared path preparation may remain a helper.

Please include table-driven coverage for `_`, `.`, `:`, Windows slash styles and 8.3 paths, plus Qwen Windows casing, and exercise the complete Qoder/Qwen `runtime cwd -> key -> session lookup` path. A real Windows canary can be supplied by maintainers if the contributor has no Windows environment.

Qwen reference: https://github.com/QwenLM/qwen-code/blob/7edc16ba11e1c4c3892c4f2664bd61d48cbf3cde/packages/core/src/utils/paths.ts

---
*Source: manual evidence review @ `b8b7911`, cross-checked against the prior Claude Windows session recovery and Qwen upstream storage code.*

zpzjzj (commented):
Re-reviewed `b8b7911` together with the earlier Claude Windows session recovery. The observed Qoder underscore fix is valid and the local/remote test gates are green, but I left one focused inline request to make workspace-key encoding a complete agent-owned contract rather than a post-transform on Claude-oriented normalization. The same contract should cover Qwen, whose upstream storage encoding has the same punctuation behavior plus Windows lowercasing.

One additional repository requirement remains: please add a concise `Unreleased / Fixed` entry to `CHANGELOG.md`. This is user-visible because affected Qoder/Qwen runs can otherwise complete grading while silently losing the native transcript and token accounting.

Expected contributor scope for this PR:
- implement the agent-owned encoder contract and wire Claude/Qoder/Qwen explicitly;
- add focused table-driven and lookup-path tests;
- update the changelog and resolve the CLA identity/signing issue.

A live Windows canary is useful but does not need to be supplied by the external contributor if no Windows environment is available; maintainers can provide that final verification.

Current status: CI, E2E, Lint, and CodeQL pass; CLA is still pending. Overall: ⚠️ conditionally mergeable after the inline design item, changelog, and CLA are addressed.

pfangueiro (approved):

PR #171, opened 2026-08-07
feat(config): support Skill include and exclude patterns

zpzjzj:
## Summary

- add optional doublestar `include` and `exclude` patterns to local Skill references
- apply the same filters to both run-agent and judge Skill installation, with `exclude` taking precedence
- keep the default behavior of installing all Skill content except `evals/`, including hidden Agent memory/metadata and ordinary YAML files
- validate patterns, preserve judge filter metadata in reports, and document the configuration in English and Chinese

For the reported Qoder RepoWiki case, users can now configure:

```yaml
skills:
  - source: local_path
    path: .
    exclude:
      - .qoder/repowiki/**
```

## Verification

- `make verify`
- `make test`
- regression coverage confirms `.yaml` and `.yamlx` have identical default handling
- installation coverage confirms `.qoder/repowiki/**` is pruned while `.claude/memory.md` remains available

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Eval-level default expect checks under `cases.defaults.expect`, with
   append-and-deduplicate semantics for list checks and case-level overrides for
   `exit_code` and `golden_file`.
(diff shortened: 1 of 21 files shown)


roark47 (approved):
LGTM

PR #173, opened 2026-08-08
feat(judge): add output_matches rule

### LINKED ISSUE #165: feat(judge): add stronger deterministic rule_based assertions
## Motivation

`skill-up` already supports useful deterministic checks in `rule_based`, such as `output_contains`, `tool_called`, `tool_called_in_turn`, and `tool_not_called_in_turn`.

For workflow-style Skill evals, a few extra assertions could make cases easier to write and more precise:

- regex-based forbidden output checks
- required / forbidden tool set checks
- lightweight structured output section checks

### 1. Regex-based forbidden output checks

```yaml
judge:
  type: rule_based
  success:
    - output_matches:
        not_regex:
          - "(?i)api[_-]?key\\s*="
          - "BEGIN PRIVATE KEY"
```

Possible semantics:

- all `not_regex` patterns must not match the final output
- invalid regex patterns should fail config validation
- use Go regexp syntax

### 2. Required / forbidden tool sets

```yaml
judge:
  type: rule_based
  success:
    - tools_required:
        names: [read_file, write_file]

    - tools_forbidden:
        names: [delete_file, send_email]
```

Possible semantics:

- `tools_required.names`: every listed tool must appear at least once in the transcript
- `tools_forbidden.names`: none of the listed tools may appear in the transcript
- this is mainly syntax sugar over multiple `tool_called` rules

### 3. Output section checks

```yaml
judge:
  type: rule_based
  success:
    - output_sections:
        required: ["Status", "Evidence", "Next Action"]
```

Possible semantics:

- each required section must appear in the final output
- the first version can stay lightweight, for example matching Markdown headings or label-style prefixes:
  - `## Status`
  - `Status:`
  - `**Status**`

## Why this fits skill-up

These additions stay within the existing deterministic `rule_based` judge model:

- no new runtime behavior
- no LLM-as-judge dependency
- no agent execution changes
- easy to cover with table-driven tests

## Open questions

1. Are these field names reasonable?
2. Should `tools_required` / `tools_forbidden` support per-turn variants in the first version?
3. How strict should `output_sections.required` be about Markdown heading syntax?

googs1025:
## Summary

  Add `output_matches` support to the deterministic `rule_based` judge so eval authors can assert final output with Go regular expressions without changing `output_contains` literal
  matching semantics.

  ## Related issues

  Closes #165

  ## Changes

  - Add `output_matches` rule schema with `all`, `any`, and `not` groups.
  - Validate configured regex patterns during config validation.
  - Evaluate `output_matches` against the final message in `RuleBasedJudge`.
  - Preserve existing failure-priority and success-all-pass semantics.
  - Add unit coverage for regex matching, empty groups, failure rules, and invalid regex validation.
  - Update user-facing and Skill author docs.

  ## Test plan

  - [x] `make test` passes
  - [ ] Manual testing steps (if applicable): N/A

  ## Notes for reviewers

  `output_contains` remains literal substring matching. Tool-set aliases and output section checks from the original issue are intentionally deferred.

### DIFF
--- a/internal/judge/rule_based.go
+++ b/internal/judge/rule_based.go
@@ -3,6 +3,7 @@ package judge
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/alibaba/skill-up/internal/config"
@@ -76,6 +77,8 @@ func evaluateAssertion(rule config.Rule, in Input) AssertionResult {
 	switch {
 	case rule.OutputContains != nil:
 		return evalOutputContains(rule.OutputContains, in.FinalMessage)
+	case rule.OutputMatches != nil:
+		return evalOutputMatches(rule.OutputMatches, in.FinalMessage)
 	case rule.ExitCode != nil:
 		return evalExitCode(*rule.ExitCode, in.ExitCode)
 	case rule.ToolCalled != nil:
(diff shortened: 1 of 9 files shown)


internal/judge/rule_based.go
```diff
@@ -174,6 +177,119 @@ func evalOutputContains(rule *config.OutputContainsRule, finalMessage string) As
...
+				Evidence: fmt.Sprintf("output does not match any of %v", rule.Any),
+			}
+		}
+	}
+
+	forbiddenPattern, invalid := firstMatchingRegex("output_matches.not", rule.Not, finalMessage)
+	if invalid != nil {
+		return *invalid
+	}
+	if forbiddenPattern != "" {
```
zpzjzj: **🟠 Major: Avoid using an empty pattern as the no-match sentinel**

`""` is a valid Go regular expression and matches every string. However, `firstMatchingRegex` returns the matched pattern and the caller treats an empty returned string as "no match". As a result, `output_matches.not: [""]` incorrectly passes for arbitrary output. I reproduced this with a focused test.

**Suggestion**: return an explicit `matched bool` from `firstMatchingRegex` (or reject empty patterns during validation), and add a regression test for an empty forbidden regex. Returning a boolean preserves the documented Go `regexp` semantics.

---
*Manual source review and focused reproduction @ 13edd02ea043d0379d2cbc4e2d35d7be9b18536a; filtered with the repository review heuristics.*

zpzjzj (commented):
## Code review summary

- 🟠 Major: 1
- 🔴 Blocker: 0

The implementation scope and the normal `all` / `any` / `not` paths look consistent with #165. One reproducible edge-case bug remains in `output_matches.not`: an empty Go regex is mistaken for the no-match sentinel. I left the details inline.

Overall: conditionally ready after the Major finding is fixed.

---
*Manual source review and focused reproduction @ `13edd02ea043d0379d2cbc4e2d35d7be9b18536a`; 1 finding accepted, 0 downgraded, 0 rejected. The configured local automated reviewer was unavailable.*

zpzjzj (approved):
LGTM

PR #180, opened 2026-08-10
fix(release): install GitHub CLI before publish

zpzjzj:
## Summary

- install a pinned GitHub CLI in the protected publish job
- verify the official Linux amd64 archive with its pinned SHA-256 checksum
- run the exact same installer on the trusted Linux x64 runner during pull requests
- verify PATH propagation and authenticated GitHub API access without publishing

## Root cause

The v0.8.0 release built, downloaded, and attested its artifacts successfully, but `Publish GitHub Release` failed with `gh: command not found`. PR #156 moved the publish job from `ubuntu-24.04` to the trusted self-hosted fleet, where GitHub CLI is not preinstalled.

The v0.8.0 release was recovered from the exact approved artifact after checksum verification. This PR prevents the same failure on future tags.

## Verification

- `make verify`
- `make test`
- `uvx zizmor@1.28.0 .github/workflows/release.yml`
- `bash -n .github/scripts/install-gh.sh`
- `Verify release tooling` runs `.github/scripts/install-gh.sh` on `[self-hosted, linux, x64, trusted]`, then checks `gh --version` and authenticated `gh api` access

### DIFF
--- a/.github/scripts/install-gh.sh
+++ b/.github/scripts/install-gh.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+gh_version="2.97.0"
+gh_linux_amd64_sha256="a2c9b8497e1f85b1ad0dfcb78b5a622e098801b8e461e459e88e1ee12f018112"
+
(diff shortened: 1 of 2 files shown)


roark47 (approved):
LGTM

PR #169, opened 2026-08-04
docs: prepare v0.8.0 release notes

zpzjzj:
## Summary
- finalize the `v0.8.0` changelog from changes merged since `v0.7.0`
- cover eval default checks, multi-iteration stability summaries, Skill include/exclude patterns, Markdown reports, the baseline CLI flag, Windows Claude token accounting, and baseline aggregate statistics
- add release links for `v0.8.0` and `v0.7.0`

## Validation
- `make verify`
- `make test`
- `git diff --check`

## Release
After this PR is merged, create tag `v0.8.0` from the resulting `main` commit. The tag triggers the release workflow.

PR #181, opened 2026-08-11
fix: resume the same CLI session across evaluation turns

roark47:
## Summary

Multi-turn evaluation never actually resumed the CLI session it started, so every
turn after the first ran in a brand-new conversation and lost all earlier
context. Turn responses were also mis-attributed: a turn could be graded against
a fragment of the previous turn, or against a `[tool: Name]` placeholder.

Reported from a real eval where turn 2 asserts on the agent remembering turn 1;
the case had to be commented out of its eval because no released version could
run it on macOS.

Root causes, all independent:

1. The session lookup rebuilt the project-directory name a CLI derives from the
   workspace path, replacing only `/`. The CLIs replace **every**
   non-alphanumeric character, so any workspace path containing `_`, `.`, a space
   or a non-ASCII character resolved to a directory that does not exist. macOS
   default `TMPDIR` paths commonly contain an underscore, which is why this
   reproduced on macOS and not on Linux `/tmp`. Without a resolved session file
   there is no session id, and `RunTurn` silently falls back to starting a new
   session — so `-r` was never sent.
2. The lookup searched the project tree unbounded and picked the newest
   `*.jsonl`. A Skill or Task call writes `<sessionID>/subagents/agent-*.jsonl`
   into that same tree; selecting one yields a file name that is not a resumable
   session id.
3. Turn boundaries were recovered by filtering on the turn numbers the session
   parser assigns, but those count transcript *events* — tool results and
   injected Skill bodies are recorded as `user` events, so one logical turn
   advances them several times.
4. Losing the session id was silent, so the only visible symptom was "the agent
   forgot the previous turn", which reads like a Skill or model problem.

## Changes

- `internal/agent/session_lookup.go`: stop reconstructing any CLI's naming rule.
  Directory names and the workspace path are compared in a shared canonical form,
  which is invariant to *which* separator a CLI substitutes, and a recorded
  working directory serves as fallback for names that cannot be derived from the
  path at all (e.g. the truncated-plus-hashed names used for long paths). Both
  steps identify the workspace explicitly, so `cases.parallelism > 1` cannot
  cross-match.
- `internal/agent/{qodercli,claude_code,qwen_code}.go`: declare the projects root
  and the depth session files live at, keeping nested transcripts out of scope.
- `internal/evaluator/multiturn.go`: a turn is the transcript appended since the
  previous turn, re-stamped with the logical turn number; warn when an engine
  reports no session id and further turns remain.
- `internal/agent/claude_code.go`: assistant messages carry only assistant text.
  Tool calls remain as their own `tool_call` entries.
- `e2e/testdata/multiturn-session/` + `e2e/multiturn_session_test.go`: a mock that
  writes real session transcripts, so the lookup and parsing path is covered end
  to end.
- `CHANGELOG.md`: entries under `Unreleased`.

## Test plan

- [x] `make test` passes (19/19 packages, `-race`)
- [x] `make verify` passes (fmt + vet + revive + golangci-lint, 0 issues)
- [x] `make e2e` passes, including the new `TestPipeline_MultiTurnSessionResume`
- [x] Manual testing steps:
  - Reproduced against the reporter's real eval with `v0.8.0` and with this
    branch, same conditions (`TMPDIR` containing an underscore, real `qodercli`,
    a PATH shim logging the argv):

    | Signal | v0.8.0 | this branch |
    | --- | --- | --- |
    | turn 2 command line | `-p …` (no `-r`) | `-r 7a08b76a-… -p …` |
    | session files for the run | 2 files, one turn each | 1 file, both turns |
    | turn 2 answer | no reference to turn 1 | carries turn 1's topic |
    | turn 1 response source | stdout fallback (lookup failed) | parsed transcript |
    | broken resume visible? | nothing in the log | n/a (resume works) |
  - The reporter confirmed the fix on their own machine, with the knowledge-base
    sub-Skill installed and the default `TMPDIR`.

## Notes for reviewers

- **Behaviour change worth calling out**: a tool call no longer appears as
  `[tool: Name]` inside assistant content. Evals asserting on that text need to
  assert on tool rules instead. One in-tree test (`ParseMCPToolCalls`) asserted
  the message count that included such a placeholder and was updated; its MCP
  assertions are unchanged.
- One pre-existing test expectation encoded the bug (`posix colon remains valid`
  asserted that `:` is *not* replaced) and was corrected.
- The Windows branch of the old key derivation is gone; the canonical comparison
  covers the cases it was added for (`C:\…` and 8.3 short names) and its
  expectations are unchanged. Not exercised on real Windows.
- Deliberately **not** included: switching the run command to `--session-id` /
  `--output-format json`, which would remove the disk lookup entirely. That is a
  larger change to how every engine is invoked and belongs in its own PR.

### DIFF
--- a/internal/agent/session_lookup.go
+++ b/internal/agent/session_lookup.go
@@ -204,9 +204,16 @@ func buildWindowsSessionLookupScript(lookup agentSessionLookup) string {
 //     that cannot be derived from the path at all, such as the truncated and
 //     hashed names the CLIs fall back to for very long paths.
 //
-// Both steps identify the workspace explicitly, so concurrent cases writing into
-// the same projects tree cannot be confused for one another. The depth bound
-// keeps nested transcripts (e.g. qodercli's <sessionID>/subagents/) out.
+// Canonicalisation is lossy in both directions: because the CLIs collapse
+// punctuation too, workspaces whose paths differ only in punctuation (say
+// /w/a_b and /w/a-b) share one project directory. Candidates are therefore
+// ranked by how well they identify the workspace: transcripts recording it as
+// their working directory first, then transcripts recording none at all (formats
+// that omit it). A transcript recording a *different* directory is never used,
+// so a colliding workspace's session cannot be resumed or graded by mistake.
+//
+// The depth bound keeps nested transcripts (e.g. qodercli's
+// <sessionID>/subagents/) out.
 //
 // Shell constraints (some runtimes provide a minimal POSIX shell): no process
 // substitution, and no pipeline whose body is shell code — such a body runs in a
@@ -227,17 +234,30 @@ func buildSessionLookupScriptWithPrinter(lookup agentSessionLookup, printBest st
 root="%s"; [ -d "$root" ] || exit 0
 ws=$(printenv %s); wskey=$(printenv %s); [ -n "$wskey" ] || exit 0
 tmp=$(mktemp) || exit 0; list=$(mktemp) || exit 0; trap 'rm -f "$tmp" "$list"' 0
+keep_recording_ws() {
+  : >"$tmp"
+  while IFS= read -r c || [ -n "$c" ]; do
+    if head -c 65536 "$c" 2>/dev/null | grep -qF "\"$ws\""; then printf '%%s\n' "$c" >>"$tmp"; fi
+  done <"$1"
+}
+add_recording_no_cwd() {
+  while IFS= read -r c || [ -n "$c" ]; do
+    if ! head -c 65536 "$c" 2>/dev/null | grep -qE '"cwd"[[:space:]]*:|"workspace-directories"'; then printf '%%s\n' "$c" >>"$tmp"; fi
+  done <"$1"
+}
 find "$root" -mindepth 1 -maxdepth 1 -type d 2>/dev/null >"$tmp" || true
 awk -v key="$wskey" '{n=$0; sub(/.*\//,"",n); gsub(/[^a-zA-Z0-9]/,"-",n); if (n==key) print}' "$tmp" >"$list" || true
 : >"$tmp"
 while IFS= read -r d || [ -n "$d" ]; do
   find "$d" -maxdepth %d -type f -name "*.jsonl"%s 2>/dev/null >>"$tmp" || true
 done <"$list"
-if [ ! -s "$tmp" ] && [ -n "$ws" ]; then
+if [ -s "$tmp" ]; then
+  cp "$tmp" "$list"
+  keep_recording_ws "$list"
+  [ -s "$tmp" ] || add_recording_no_cwd "$list"
+elif [ -n "$ws" ]; then
   find "$root" -maxdepth %d -type f -name "*.jsonl"%s 2>/dev/null >"$list" || true
-  while IFS= read -r c || [ -n "$c" ]; do
-    if head -c 65536 "$c" 2>/dev/null | grep -qF "\"$ws\""; then printf '%%s\n' "$c" >>"$tmp"; fi
-  done <"$list"
+  keep_recording_ws "$list"
 fi
 best=; ts=-1
 while IFS= read -r p || [ -n "$p" ]; do
--- a/internal/agent/session_lookup.go
+++ b/internal/agent/session_lookup.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -172,12 +173,20 @@ func canonicalWorkspaceKey(workspace string) string {
 // matched in the escaped form it is actually written in ("C:\\Users\\..."), and
 // the key is part of the fragment, so a path that merely appears somewhere in the
 // conversation — a tool argument, say — is not mistaken for recorded identity.
+//
+// HTML escaping is disabled to match the producer: the CLIs are JavaScript
+// programs, and JSON.stringify writes "&", "<" and ">" literally where Go's
+// default encoder emits \u0026, \u003c and \u003e. A path containing any of them
+// would otherwise never match. (Go still escapes U+2028/U+2029 where JavaScript
+// does not; those are line separators and not a realistic path character.)
 func workspaceCwdMarkers(workspace string) []string {
-	encoded, err := json.Marshal(workspace)
-	if err != nil {
+	var buf bytes.Buffer
+	encoder := json.NewEncoder(&buf)
+	encoder.SetEscapeHTML(false)
+	if err := encoder.Encode(workspace); err != nil {
 		return nil
 	}
-	quoted := string(encoded)
+	quoted := strings.TrimRight(buf.String(), "\n")
 
 	markers := make([]string, 0, 4)
 	for _, key := range []string{`"cwd":`, `"directories":[`} {
@@ -197,10 +206,26 @@ func extractSessionIDFromPath(sessionPath string) string {
 	return base
 }
 
+// pathComparison captures how paths are compared on the target platform: POSIX
+// paths are case-sensitive, Windows paths are not, and the same directory may
+// reach us spelled differently from how the CLI recorded it.
+type pathComparison struct {
+	// awkNameMatch is the awk condition comparing a canonicalised directory name
+	// (n) against the workspace key (key).
+	awkNameMatch string
+	// grepFlags selects fixed-string matching of the recorded-cwd markers.
+	grepFlags string
+}
+
+var (
+	caseSensitivePaths   = pathComparison{awkNameMatch: "n==key", grepFlags: "-qFf"}
+	caseInsensitivePaths = pathComparison{awkNameMatch: "tolower(n)==tolower(key)", grepFlags: "-qiFf"}
+)
+
 // buildSessionLookupScript renders the shared shell snippet that picks the
 // newest *.jsonl under the configured root.
 func buildSessionLookupScript(lookup agentSessionLookup) string {
-	return buildSessionLookupScriptWithPrinter(lookup, `printf %s "$best"`)
+	return buildSessionLookupScriptWithPrinter(lookup, `printf %s "$best"`, caseSensitivePaths)
 }
 
 func buildWindowsSessionLookupScript(lookup agentSessionLookup) string {
@@ -210,6 +235,7 @@ func buildWindowsSessionLookupScript(lookup agentSessionLookup) string {
 	return buildSessionLookupScriptWithPrinter(
 		lookup,
 		`if [ -n "$best" ]; then cygpath -w "$best" 2>/dev/null || printf %s "$best"; fi`,
+		caseInsensitivePaths,
 	)
 }
 
@@ -246,6 +272,11 @@ func buildWindowsSessionLookupScript(lookup agentSessionLookup) string {
 // `--session-id`, and a session_id field under `--output-format json`), which is a
 // change to how every engine is invoked and is tracked separately.
 //
+// Path comparison follows the platform: POSIX paths are case-sensitive, while on
+// Windows the same directory may be spelled with different case by the runtime
+// and by the CLI, so both the directory match and the recorded working directory
+// are folded there.
+//
 // The depth bound keeps nested transcripts (e.g. qodercli's
 // <sessionID>/subagents/) out.
 //
@@ -254,7 +285,7 @@ func buildWindowsSessionLookupScript(lookup agentSessionLookup) string {
 // subshell, and a subshell exiting can fire the EXIT trap and delete the temp
 // files the outer script still needs. Candidate lists therefore go through temp
 // files that plain `while ... done <file` loops read back.
-func buildSessionLookupScriptWithPrinter(lookup agentSessionLookup, printBest string) string {
+func buildSessionLookupScriptWithPrinter(lookup agentSessionLookup, printBest string, paths pathComparison) string {
 	extra := ""
 	if lookup.findExtra != "" {
 		extra = " " + lookup.findExtra
@@ -277,7 +308,7 @@ keep_recording_ws() {
   : >"$tmp"
   [ -s "$marks" ] || return 0
   while IFS= read -r c || [ -n "$c" ]; do
-    if head -c 65536 "$c" 2>/dev/null | grep -qFf "$marks"; then printf '%%s\n' "$c" >>"$tmp"; fi
+    if head -c 65536 "$c" 2>/dev/null | grep %s "$marks"; then printf '%%s\n' "$c" >>"$tmp"; fi
   done <"$1"
 }
 any_records_cwd() {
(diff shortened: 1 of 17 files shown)


roark47:
Addressed in 3db0de2 — see the [reply in the review thread](https://github.com/alibaba/skill-up/pull/181#discussion_r3756250483) for the reasoning, the ranking the lookup now applies, and the regression test.

(Superseding an earlier revision of this comment that cited a commit hash from before a rebase onto the updated branch.)

internal/agent/session_lookup.go
```diff
@@ -161,21 +189,67 @@ func buildWindowsSessionLookupScript(lookup agentSessionLookup) string {
...
+	}
+
 	return fmt.Sprintf(`home=$(printenv HOME); [ -n "$home" ] || exit 0
 root="%s"; [ -d "$root" ] || exit 0
-tmp=$(mktemp) || exit 0; trap 'rm -f "$tmp"' 0
-find "$root" -type f -name "*.jsonl"%s 2>/dev/null >"$tmp" || true
+ws=$(printenv %s); wskey=$(printenv %s); [ -n "$wskey" ] || exit 0
+tmp=$(mktemp) || exit 0; list=$(mktemp) || exit 0; trap 'rm -f "$tmp" "$list"' 0
+find "$root" -mindepth 1 -maxdepth 1 -type d 2>/dev/null >"$tmp" || true
+awk -v key="$wskey" '{n=$0; sub(/.*\//,"",n); gsub(/[^a-zA-Z0-9]/,"-",n); if (n==key) print}' "$tmp" >"$list" || true
```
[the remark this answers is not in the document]
roark47: Valid, and the collision is one level deeper than described.

The CLIs collapse punctuation when they name a project directory, so `/w/a_b` and `/w/a-b` do not merely canonicalize to the same key on our side — the CLI itself stores both workspaces' sessions **in the same directory**. The ambiguity is therefore not "several matching directories" but "one directory holding several workspaces' sessions", which no directory-name rule can separate. Requiring an unambiguous directory match would not help; the recorded working directory is the only usable discriminator.

Worth noting this is a risk the PR introduced: the previous key kept `_`, so the lookup pointed at a directory the CLI never writes and simply found nothing. Making it match turned "finds nothing" into "may pick the wrong one", which is the worse failure of the two — it silently resumes another conversation and grades against it.

Fixed in 3db0de2 by ranking candidates instead of taking the newest outright:

1. transcripts recording this workspace as their working directory;
2. otherwise transcripts recording none at all, which keeps formats that omit it (qwen's chat JSONL) working as before;
3. a transcript recording a *different* working directory is never used.

The wrong session is no longer reachable. The residual failure mode is losing the lookup, which now emits a warning instead of degrading silently, so it surfaces as a broken session rather than as a Skill that "forgot" the previous turn.

`TestFindQoderSessionFileDisambiguatesCollidingWorkspaces` covers it: it builds the colliding pair `ws_x` / `ws-x` (asserting they really do collide, so the fixture cannot rot into a no-op), makes the *other* workspace's transcript the newest file, and runs both directions — our transcript recording its cwd, and our transcript recording none while the other one does. Both subtests failed before the change and pass after.

internal/agent/session_lookup.go
```diff
@@ -161,21 +189,87 @@ func buildWindowsSessionLookupScript(lookup agentSessionLookup) string {
...
+: >"$tmp"
+while IFS= read -r d || [ -n "$d" ]; do
+  find "$d" -maxdepth %d -type f -name "*.jsonl"%s 2>/dev/null >>"$tmp" || true
+done <"$list"
+if [ -s "$tmp" ]; then
+  cp "$tmp" "$list"
+  keep_recording_ws "$list"
+  [ -s "$tmp" ] || add_recording_no_cwd "$list"
```
zpzjzj: **🔴 Blocker: Do not infer workspace identity from transcript-wide substring/no-cwd fallback**

`keep_recording_ws` searches any JSON string equal to `$ws`, not the `cwd` metadata specifically, and this fallback admits every candidate with no cwd before choosing the newest by mtime. Therefore the invariant that another workspace's session is unreachable still does not hold:

- Two canonical-colliding workspaces whose transcripts both omit cwd remain indistinguishable; Qwen session events have no cwd, so the newer other-case transcript can be parsed and graded.
- An unrelated transcript whose user/tool content is exactly the target workspace path passes the raw `grep`.
- A real Windows cwd is JSON-escaped (`C:\\...`) and does not match raw `C:\...`, so the correct transcript is rejected and resume loses its ID.

The tests only cover “ours omits cwd, other records it”; their fixtures concatenate unescaped paths, and shell lookup is skipped on Windows.

**Suggestion:** match only JSON-encoded, explicit `cwd`/`workspace-directories` metadata. For formats without identity metadata, use run-scoped/CLI session identity or fail safely on an ambiguous collision instead of selecting the newest file. Add both-no-cwd, content-false-positive, and JSON-escaped Windows tests.

---
*Manual source review and focused predicate reproduction at `3db0de2`; filtered with repository review heuristics.*

zpzjzj (commented):
## Code review summary

- 🔴 Blocker: 1
- 🟠 Major: 0

The latest collision fix improves one case, but the shared lookup can still select another workspace's transcript through no-cwd candidates or transcript-wide raw matching, and it rejects JSON-escaped Windows cwd values. Details are in the inline comment.

Overall: not ready to merge until workspace identity is made unambiguous.

---
*Manual source review and focused reproduction at `3db0de2`; 1 accepted, 0 downgraded, 0 rejected. Local Claude reviewer unavailable.*

internal/agent/session_lookup.go
roark47: All three points are correct, and the second one is a regression this PR introduced rather than a pre-existing gap. Fixed in 2d5127c.

**Content false positives.** Confirmed: a transcript that passes our workspace path to a tool contains it as a plain JSON string, indistinguishable from a recorded cwd under a raw match. Worth noting the first version of my regression test did *not* reproduce this — it embedded the path inside an escaped string (`\"…\"`), which the raw grep happened to miss. The test now uses a tool argument, which is both realistic and actually red before the fix.

**JSON-escaped Windows cwd.** Confirmed, and worse than "the correct transcript is rejected": with the rejection came the no-cwd exclusion, so a Windows Claude run resolved to *nothing* — losing exactly the token usage that #169 restored. This is the most serious of the three and it was not covered anywhere, since the shell lookup is skipped on Windows in CI.

**Both-no-cwd.** Correct that it stays unresolved by directory matching, and your framing pushed me to separate detectable from undetectable ambiguity.

The predicate is now the JSON fragment a transcript is actually written as — `"cwd":<encoded path>` / `"directories":[<encoded path>` — built with `encoding/json`, so escaping matches per platform, and the key is part of the pattern. Candidates resolve in order:

1. transcripts recording this workspace;
2. otherwise **nothing**, if any candidate records a *different* working directory — that proves the directory is shared, and reporting a lost session (which the evaluator now warns about) beats guessing;
3. otherwise the newest, which is the only option for formats that never record one.

I did not make step 3 unconditional-fail-safe, and that is a deliberate trade-off: qwen's chat JSONL carries no working directory at all, and neither did older CLI releases, so refusing unattributable transcripts outright would drop transcript and token reporting for every ordinary qwen run — not only colliding ones. Two colliding workspaces whose transcripts all omit identity therefore remain indistinguishable *on disk*; the honest fix for that is run-scoped identity, i.e. passing `--session-id` and reading `session_id` back from `--output-format json` so the filesystem is never consulted for identity. That changes how every engine is invoked and is deliberately kept out of this PR; I can open the follow-up issue and reference it here if you prefer that tracked before merge.

Tests added, each red before the change:

- `TestFindQoderSessionFileIgnoresPathMentionedInContent` — colliding neighbour passes our path to a tool and is the newest file.
- `TestFindClaudeSessionFileMatchesJSONEscapedWindowsCwd` — Windows-style workspace with `"cwd":"C:\\Users\\…"`, plus a colliding `skill-up-1` vs `skill_up-1` neighbour; runs under a POSIX shell so it is not skipped on CI.
- `TestFindQoderSessionFileFailsSafeWhenIdentityUnknown` — our transcript omits cwd while the neighbour's records its own; expects no result.

`TestFindQoderSessionFile_SelectsNewestByModTime` (fixtures without cwd) is what pins step 3, and it caught an over-strict intermediate version of this change.

Verified beyond unit tests: `make verify` clean, 19/19 packages, `make e2e` green, and the reporter's real two-turn eval re-run end to end against real `qodercli` — `-r 4e3ada64-…` sent, both turns in one session file, turn 2 carrying turn 1's topic.

roark47:
### Note on the CodeQL failure (and one out-of-scope commit)

CodeQL reported `go/allocation-size-overflow` (high) at `internal/runtime/runtime.go:55`:

```go
env := make(map[string]string, len(persistentEnv)+len(callEnv))
```

That code is **not** part of this PR's work — `git diff origin/main...HEAD -- internal/runtime/` is empty, and the expression dates back to `47c6b4a chore: initial public release`. It surfaced here only because this branch merged `main` and the PR scan diffs against the merge result. The rule already has a long-standing open alert on `main` for the same pattern in `internal/userconfig/apply.go:105`.

I fixed it in place anyway (9a450c9) rather than asking for a dismissal, so the branch reports clean and nobody has to decide whether a red check is real. The hint now uses the larger of the two lengths: keys overlap between the maps, so it is a sound lower bound with no arithmetic feeding the allocation, and a comment records why the sum is not restored later.

Two things to flag explicitly:

- This is one commit outside the PR's stated scope. If you would rather keep the branch focused, say so and I will pull it out into its own PR and let this one merge with the finding dismissed as pre-existing.
- The sibling alert in `internal/userconfig/apply.go:105` is untouched, since it does not block this branch. Happy to fix it in the same follow-up.

internal/agent/session_lookup.go
```diff
@@ -119,19 +138,52 @@ func findAgentSessionJSONL(ctx context.Context, rt Runtime, lookup agentSessionL
...
+	return nonAlphanumeric.ReplaceAllString(workspace, "-")
+}
+
+// workspaceCwdMarkers renders the JSON fragments that mark a transcript as this
+// workspace's. The path is JSON-encoded, so a Windows working directory is
+// matched in the escaped form it is actually written in ("C:\\Users\\..."), and
+// the key is part of the fragment, so a path that merely appears somewhere in the
+// conversation — a tool argument, say — is not mistaken for recorded identity.
+func workspaceCwdMarkers(workspace string) []string {
+	encoded, err := json.Marshal(workspace)
```
zpzjzj: **🟠 Major: Preserve Windows path equivalence when generating cwd markers**

`json.Marshal(workspace)` does not necessarily produce the spelling used by the Node-based CLIs: Go HTML-escapes a valid path such as `C:\\Users\\A&B\\...` to contain `A\u0026B`, while `JSON.stringify` writes `A&B`. The fixed-string marker therefore rejects the correct transcript. The directory-key and marker comparisons are also case-sensitive even though Windows paths are case-insensitive.

I reproduced both cases at `9a450c9`: a transcript for a workspace containing `&`, and one whose recorded cwd differs only as `runneradmin` vs `RUNNERADMIN`, each made `findClaudeSessionFile` return an empty path. Downstream this loses the native transcript/token counts and leaves Qoder without the session ID needed to resume the next turn. The former is a regression from the old direct project-key lookup, which did not require the cwd marker after locating the directory.

**Suggestion:** encode markers with the same JSON spelling as the producer (at minimum `SetEscapeHTML(false)`), and apply Windows-specific case-insensitive normalization to both the project key and recorded cwd comparison—ideally decode the explicit cwd field and compare it using Windows path semantics. Add regression cases for `&` and case-only differences, then validate the full lookup/download path on a Windows runner.

---
*Manual source review and focused reproductions at `9a450c9`; filtered with repository review heuristics.*

zpzjzj (commented):
## Follow-up code review summary

- 🔴 Blocker: 0 new
- 🟠 Major: 1 new

The previous raw-path/JSON-escaping fix improves the common Windows case, but workspace identity still fails for valid paths containing `&` and for Windows-equivalent paths that differ only in case. The focused reproductions and suggested fix are in the inline comment.

Windows validation is still incomplete: the new session lookup tests execute under a POSIX/Linux runtime descriptor, the multi-turn E2E skips Windows, and the required PR checks are Linux-only. After fixing the marker semantics, please run the live Windows Claude lookup-to-token test at this exact head; ideally add a Windows multi-turn resume path as well.

The already-discussed both-no-cwd/Qwen ambiguity is not repeated here.

Overall: conditionally mergeable after the Windows identity regression is fixed and the real Windows chain is validated.

---
*Manual source review and focused reproductions at `9a450c9`; 1 accepted, 0 downgraded, 1 historical issue deduplicated. Local Claude reviewer unavailable.*

internal/agent/session_lookup.go
roark47: Both points are correct and fixed in 71e9305. The Windows validation request I cannot honestly satisfy — details at the end rather than buried.

**HTML escaping.** Confirmed, and the framing understates it: this is not Windows-specific. Measured at `9a450c9`:

```
path                 Go json.Marshal              SetEscapeHTML(false) / JSON.stringify
C:\Users\A&B\ws      "C:\\Users\\A\u0026B\\ws"    "C:\\Users\\A&B\\ws"
/tmp/a<b>c           "/tmp/a\u003cb\u003ec"       "/tmp/a<b>c"
```

`&`, `<` and `>` are all affected, so `/tmp/a&b` on Linux or macOS fails exactly like the Windows example. Markers are now built with `SetEscapeHTML(false)`, which matches `JSON.stringify` byte for byte on those inputs and on non-ASCII. One residual difference is documented in the code: Go escapes U+2028/U+2029 where JavaScript does not — line separators, not a realistic path character, and I preferred noting it over hand-rolling a JSON string encoder.

**Windows case equivalence.** Confirmed for both comparisons. The project-directory match now folds case in awk and the marker match uses `grep -i`, but only in the Windows script variant: POSIX paths are case-sensitive and `/w/ws_a` must stay distinct from `/w/WS_A`, so that is pinned by `TestFindQoderSessionFileKeepsPOSIXPathCase` to stop the folding from leaking. The platform difference is passed as a `pathComparison` value rather than a boolean flag.

Also worth agreeing with explicitly: "a regression from the old direct project-key lookup, which did not require the cwd marker after locating the directory" is exactly right, and it is why marker correctness now carries weight it did not before. I did not add a "marker failed, take the newest anyway" safety valve, since that would reinstate the wrong-session risk from your earlier blocker; the answer has to be that marker matching is right, plus step 3 for formats that record nothing at all.

Tests added, each red before the change:

- `TestFindQoderSessionFileWithHTMLSensitiveWorkspace` — workspace `ws&a<b>c`, POSIX, so it runs on the existing CI.
- `TestFindClaudeSessionFileFoldsWindowsPathCase` — `RunnerAdmin` from the runtime vs `RUNNERADMIN` recorded, with the project directory created under the CLI's spelling.
- `TestFindQoderSessionFileKeepsPOSIXPathCase` — the counterpart guard described above.

**On validating the real Windows chain: I cannot do that, and I would rather say so than imply otherwise.** I have no Windows machine or runner access here; everything above was verified on macOS. What the new tests do cover is the Windows *code path* — they drive a runtime descriptor reporting `GOOS=windows`, so the Windows script variant (case folding plus the `cygpath` printer) is what executes, under bash. What they cannot cover is real Windows behaviour: actual `cygpath` output, Git Bash path spellings, and what Claude Code genuinely writes into `%USERPROFILE%\.claude\projects` on that platform.

So the live lookup-to-token test at this head is still outstanding, and the required checks remain Linux-only. Two ways forward, your call:

1. You or someone with a Windows environment runs it at `71e9305` and I fix whatever falls out.
2. I add a Windows job to CI covering the session-lookup tests (and, if the runner has Claude available, the multi-turn resume path) — a CI change I would rather propose separately than fold into this PR.

Until one of those happens, I would describe the Windows side of this PR as "reasoned and unit-covered, not field-verified", and it is fair to hold merge on that if you want the guarantee.

zpzjzj (approved):
LGTM

PR #183, opened 2026-08-12
feat(compare): add offline result comparison

### LINKED ISSUE #175: feat(cli): add offline result compare command
## Background

Now skill-up can run evals, generate reports, and create baseline benchmark data. But I think it still misses one offline command to compare two finished runs.

When we change a Skill, review a PR, or upgrade model, we often want to know:

- which cases are fixed
- which cases are regressed
- how pass rate changes
- how tokens and duration change
- whether engine or model is different
- whether this can be used as a CI gate

## Proposal

Add a new command:

```bash
skill-up compare <old-result.json> <new-result.json>
```

It should compare two existing result/report JSON files. It should not run eval again, and should not call any agent or judge.

## Scope

The command can:

- read JSON files compatible with current report.Input/result.json format
- use primary case result semantics, so benchmark without_skill entries are not counted twice
- show run level delta:
  - pass rate
  - case counts
  - total/input/output tokens
  - duration
- show case level transition:
  - fixed
  - regressed
  - unchanged
  - added
  - removed
- show metadata difference, like skill name, engine, model, schema version and time
- support default text output
- support --format json for CI or other tools
- support optional gates, for example fail on regression or too much token increase

## Non-goals

- Do not rerun evals
- Do not call agent or judge
- Do not add dashboard/cloud/snapshot feature in this issue
- First version does not need markdown output

## Acceptance criteria

- `skill-up compare old.json new.json` prints a readable summary
- `--format json` prints stable machine readable output
- benchmark result is not double counted
- fixed/regressed/added/removed cases are classified correctly
- configured gate failure returns non-zero exit code
- invalid input gives clear error message

googs1025:
## Summary
- Add offline `skill-up compare <old-result.json> <new-result.json>` for text and JSON result diffs.
- Compare primary case metrics, metadata, case transitions, and CI gates without rerunning evals.
- Document the new command and update changelog.

## Test Plan
- `GOCACHE=/private/tmp/skill-up-go-build-cache GOTOOLCHAIN=go1.26.4 make verify`
- `GOCACHE=/private/tmp/skill-up-go-build-cache GOTOOLCHAIN=go1.26.4 make test`

Closes #175

### DIFF
--- a/internal/compare/compare.go
+++ b/internal/compare/compare.go
@@ -1,0 +1,201 @@
+// Package compare compares two offline skill-up report results.
+package compare
+
+import (
+	"time"
+
+	"github.com/alibaba/skill-up/internal/judge"
+	"github.com/alibaba/skill-up/internal/report"
+)
+
+// Options controls compare output and gate evaluation.
+type Options struct {
+	FailOnRegression        bool
+	MaxRegressions          *int
+	MaxTokenIncreasePercent *float64
+}
+
+// Result is the stable comparison result used by text and JSON output.
+type Result struct {
+	Metadata MetadataDiff    `json:"metadata"`
+	Run      RunComparison   `json:"run"`
+	Cases    CaseTransitions `json:"cases"`
+	Gates    GateResult      `json:"gates"`
+}
+
+// MetadataDiff records old and new run metadata.
+type MetadataDiff struct {
+	SkillName     FieldDiff[string]    `json:"skill_name"`
+	SchemaVersion FieldDiff[string]    `json:"schema_version"`
+	EngineName    FieldDiff[string]    `json:"engine_name"`
+	ModelName     FieldDiff[string]    `json:"model_name"`
+	StartTime     FieldDiff[time.Time] `json:"start_time"`
+	EndTime       FieldDiff[time.Time] `json:"end_time"`
+}
+
+// FieldDiff stores old/new values and whether they differ.
+type FieldDiff[T comparable] struct {
+	Old     T    `json:"old"`
+	New     T    `json:"new"`
+	Changed bool `json:"changed"`
+}
+
+// RunComparison stores old/new/delta aggregate metrics.
+type RunComparison struct {
+	Old   RunMetrics `json:"old"`
+	New   RunMetrics `json:"new"`
+	Delta RunMetrics `json:"delta"`
+}
+
+// RunMetrics stores aggregate metrics for one result set or a delta.
+type RunMetrics struct {
+	CaseCount    int     `json:"case_count"`
+	PassCount    int     `json:"pass_count"`
+	PassRate     float64 `json:"pass_rate"`
+	TotalTokens  int     `json:"total_tokens"`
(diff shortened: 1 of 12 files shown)


zpzjzj:
Thanks for the proposal and implementation. I understand that this command provides domain-aware comparison, such as primary-case handling, case transition classification, and CI exit codes.

However, if the scope is only comparing two local `result.json` files, the value of adding a dedicated built-in command is not yet very clear to me. The current pass-rate, status, token, and duration comparison can already be implemented fairly cheaply with `jq` or a small script, while this PR introduces a relatively large amount of code and a new CLI contract that we will need to maintain.

I think the feature would be more compelling if it formed part of a complete baseline workflow—for example:

- locating or selecting the baseline result automatically;
- comparing against the latest release or a named baseline;
- handling result schema compatibility across versions;
- providing richer regression evidence beyond case status;
- offering a reusable CI integration that individual repositories would otherwise need to maintain themselves.

Could you clarify the concrete user workflow this is intended to unlock, and what advantage the built-in command provides over a documented `jq`/script example? If the intention is only raw offline file comparison, I’m not yet convinced that the maintenance cost is justified.

internal/compare/compare.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	result.Gates = EvaluateGates(result, options)
+	return result
+}
+
+func collectRunMetrics(input report.Input) RunMetrics {
+	primary := input.PrimaryCaseResults()
+	metrics := RunMetrics{
+		CaseCount:   len(primary),
+		TotalTokens: input.TotalTokens,
```
JHWang-1997, line 107: **Use primary-case tokens for the token growth gate**

This field uses the report-level `TotalTokens`, while `InputTokens` and `OutputTokens` below are calculated from `PrimaryCaseResults()`. In benchmark mode, report-level `TotalTokens` includes both `with_skill` and `without_skill` executions, so changes in the baseline can mask a token regression in the tested skill.

For example:

| Run | `with_skill` tokens | `without_skill` tokens | report `TotalTokens` |
| --- | ---: | ---: | ---: |
| Old | 100 | 100 | 200 |
| New | 120 | 50 | 170 |

The tested skill's primary token usage increased by 20%, but `--max-token-increase-percent 10` sees `200 -> 170` and incorrectly passes because the full benchmark total decreased by 15%. The text output is also internally inconsistent: primary input/output tokens increase while `total tokens` decreases.

Since this command otherwise applies primary-case semantics to run metrics and transitions, please derive the comparable total from the primary results as well, for example after accumulation:

```go
metrics.TotalTokens = metrics.InputTokens + metrics.OutputTokens
```

If full-run benchmark cost is also useful, it should be exposed under a separate, clearly named metric and should not drive the primary token regression gate.

JHWang-1997 (commented):
The compare command is well structured, but the token gate mixes primary-case metrics with full benchmark-run totals. This can make CI report the opposite of the tested skill's actual token trend. Please see the inline example and suggested fix.

PR #184, opened 2026-08-13
fix(agent): handle large Codex JSONL records

### LINKED ISSUE #172: [Bug] Codex JSONL parser silently truncates output when a line exceeds 1 MiB
## Description

When the Codex engine emits a JSONL record larger than 1 MiB, `skill-up` silently stops parsing the output and retains an earlier assistant message as `SessionResult.FinalMessage`.

For an `agent_judge` run, this can produce a misleading error:

```text
agent_judge failed to parse agent output:
no valid JSON found in agent output (length=...)
```

The Judge may have produced valid final JSON later in the same `stdout.json`, but `skill-up` never reaches it.

This is especially easy to trigger when the Judge inspects a large transcript with `rg`, `grep`, or another command that writes many matches into one `command_execution.aggregated_output` field.

## Steps to reproduce

1. Configure an evaluation using:

   ```yaml
   engine:
     name: codex
   ```

   and an `agent_judge`.

2. During the Judge run, execute a command whose output is large enough for one Codex JSONL record to exceed 1,048,576 bytes.

   For example, search a large transcript without limiting the number or size of matching lines.

3. Let the Judge emit a valid final JSON response after that command:

   ```json
   {
     "results": [
       {
         "criterion": "example criterion",
         "passed": true,
         "evidence": "example evidence"
       }
     ]
   }
   ```

4. Run the evaluation:

   ```bash
   skill-up run evals/eval.yaml
   ```

5. Inspect:

   ```text
   <report>/<case>/with_skill/outputs/judge/run/stdout.json
   ```

6. Compare:

   - the first JSONL line larger than 1 MiB;
   - the last assistant message before that line;
   - the valid final `agent_message` after that line;
   - the length reported by `no valid JSON found`.

## Expected behavior

One of the following should happen:

1. `skill-up` successfully reads the complete Codex output and uses the final Judge JSON; or
2. `skill-up` returns an explicit output-scanning error identifying the oversized JSONL record.

If `--output-last-message` produced a valid last-message artifact, that artifact should be used as the authoritative final response instead of an earlier progress message parsed from incomplete stdout.

## Actual behavior

`skill-up` stops scanning at the first JSONL line larger than 1 MiB.

The scan error is not returned or logged. The parser keeps the last assistant progress message seen before the oversized line and treats it as the final response.

`agent_judge` then attempts to parse that progress message as JSON and reports:

```text
no valid JSON found in agent output
```

This makes a valid Judge result appear to be an evaluation infrastructure failure.

## Observed evidence

I reproduced this in two independent Judge runs.

### Run A

- First oversized physical JSONL line: line 20
- Line size: 1,117,390 UTF-8 bytes
- Record type: completed `command_execution`
- `aggregated_output` size: 942,696 characters
- Valid final Judge JSON: line 35
- Final Judge JSON size: 4,780 characters
- Last assistant message before the oversized line: line 17
- That earlier message size: exactly 199 UTF-8 bytes
- Reported error:

  ```text
  no valid JSON found in agent output (length=199)
  ```

### Run B

- Oversized physical JSONL lines:
  - line 16: 1,142,738 UTF-8 bytes
  - line 17: 1,140,287 UTF-8 bytes
- Corresponding `aggregated_output` sizes:
  - 1,007,299 characters
  - 983,543 characters
- Valid final Judge JSON: line 36
- Final Judge JSON size: 6,867 characters
- Last assistant message before the first oversized line: line 12
- That earlier message size: exactly 243 UTF-8 bytes
- Reported error:

  ```text
  no valid JSON found in agent output (length=243)
  ```

All physical records in both `stdout.json` files are valid JSON when read with a parser that supports lines larger than 1 MiB.

The final Judge `agent_message` in both runs is also valid JSON.

As a control, another run whose largest JSONL line was 954,948 bytes completed Judge parsing normally.

## Suspected root cause

At commit [`c6ca37aeb6c36eca1141101d0632401414957705`](https://github.com/alibaba/skill-up/commit/c6ca37aeb6c36eca1141101d0632401414957705), `parseCodexOutput` uses a `bufio.Scanner` with a hard 1 MiB maximum:

```go
scanner := bufio.NewScanner(strings.NewReader(output))
const maxTokenLen = 1024 * 1024
buf := make([]byte, maxTokenLen)
scanner.Buffer(buf, maxTokenLen)

for scanner.Scan() {
    // ...
}
```

Source:

https://github.com/alibaba/skill-up/blob/c6ca37aeb6c36eca1141101d0632401414957705/internal/agent/codex.go#L683-L704

There is no `scanner.Err()` check after the scan loop.

`parseCodexSessionFile` has the same 1 MiB limit and also does not propagate the scanner error:

https://github.com/alibaba/skill-up/blob/c6ca37aeb6c36eca1141101d0632401414957705/internal/agent/codex.go#L929-L978

There is an additional fallback issue in `resolveCodexLastMessage`:

```go
if finalMsg != "" || lastMessagePath == "" {
    return finalMsg, trans, generatedFiles, func() {}
}
```

Source:

https://github.com/alibaba/skill-up/blob/c6ca37aeb6c36eca1141101d0632401414957705/internal/agent/codex.go#L508-L532

Because an earlier progress message was successfully parsed before the oversized line, `finalMsg` is non-empty. Therefore, the valid file produced through `--output-last-message` is not read.

## Suggested fix

Please consider addressing all three parts:

1. Detect and propagate `scanner.Err()` from both:
   - `parseCodexOutput`
   - `parseCodexSessionFile`

2. Increase or remove the fixed 1 MiB physical-line limit.

   A streaming approach that supports arbitrarily large JSONL records would be safer than relying on another fixed upper bound.

3. Prefer the `--output-last-message` artifact as the authoritative final message when it exists, even when stdout parsing found an earlier assistant message.

   Alternatively, only trust the stdout-derived final message when parsing reached a complete terminal event without a scanner error.

## Suggested regression test

Add a Codex parser test containing:

1. an assistant progress message;
2. a valid `command_execution` JSONL record larger than 1 MiB;
3. a valid final `agent_message` containing Judge JSON;
4. a terminal `turn.completed` event.

The expected result should assert that:

- the oversized record does not silently truncate parsing;
- the valid final message is returned;
- scanner errors are never ignored;
- `agent_judge` receives the final JSON rather than the earlier progress message.

A second test should verify that `--output-last-message` recovers the correct final response when stdout parsing is incomplete.

## Environment

- skill-up version: `skill-up version 0.7.0`
- Go version: `go1.26.5 windows/amd64`
- OS: `Microsoft Windows NT 10.0.26200.0`
- Architecture: `X64`
- PowerShell: `5.1.26100.8875`
- Engine: Codex

## Additional context

The current `main` branch still has the same `internal/agent/codex.go` and `internal/judge/agent_judge.go` file contents as release `v0.7.0`, so updating from v0.7.0 to the current main branch does not appear to avoid this failure.

zpzjzj:
## Summary

- capture Codex JSONL outside the evaluated workspace so large output is not buffered by `Runtime.Exec`
- enforce configurable 16 MiB per-record and 256 MiB total-output ceilings with explicit parse errors
- stream runtime downloads, propagate session read failures, and preserve authoritative `--output-last-message` results
- use target-runtime path semantics and transfer-specific timeouts for native and remote runtimes
- add regression coverage for large records, output ceilings, artifact isolation, and cleanup behavior

## Validation

- `make verify`
- `make test`
- `make build`
- `codex review --commit 0babba7` (no actionable findings; `5d8d93e` only shortens the function signature to satisfy CI funlen)

Closes #172


internal/agent/codex.go
```diff
@@ -924,71 +930,71 @@ type codexSessionParseResult struct {
...
-	const maxTokenLen = 1024 * 1024
-	buf := make([]byte, maxTokenLen)
-	scanner.Buffer(buf, maxTokenLen)
-
-	for scanner.Scan() {
-		line := strings.TrimSpace(scanner.Text())
+	reader := bufio.NewReader(file)
+	var readErr error
+	for {
+		rawLine, err := reader.ReadString('\n')
```
roark47: [P1] Keep a hard ceiling for untrusted JSONL records

`ReadString` buffers until a newline with no size limit, and the stdout path now accepts an unlimited physical line as well. These records embed command/MCP output derived from the evaluated repository, so a crafted repository can make Codex emit a multi-gigabyte single record and OOM the `skill-up` host while this code and `json.Unmarshal` allocate it. None of the runtimes currently caps captured stdout or downloaded file size, so replacing the 1 MiB ceiling with infinity introduces a process-level denial of service. Please use a higher/configurable per-record ceiling and return the explicit parsing error requested in #172 when it is exceeded.

roark47 (commented):
Security/robustness review: one high-severity availability issue found.

internal/agent/codex.go
zpzjzj: Fixed in 0babba7. Codex JSONL now has a configurable 16 MiB per-record ceiling (`max_jsonl_record_bytes`) and a 256 MiB total-artifact ceiling (`max_jsonl_output_bytes`), with explicit errors when either is exceeded. Stdout is redirected to an isolated framework temp path so `Runtime.Exec` does not buffer it or mutate the evaluated workspace, downloads are streamed with transfer-specific timeouts, and both stdout/session parsers share the bounded record reader. Added regression tests for >1 MiB valid records, exact/over-limit records, total-size rejection, remote-path semantics, and final-message recovery; `make verify`, `make test`, `make build`, and final Codex self-review all pass.

roark47 (approved):

PR #185, opened 2026-08-13
docs(skill-upper): add YAML field comments

### LINKED ISSUE #174: Add simple field comments to generated eval YAML files
## Problem / motivation

When using the `skill-upper` skill to generate `eval.yaml` or case YAML files, the output YAML is valid, but some fields are not easy to understand for new users.

This is harder for users who are new to skill-up, or users who are not native English speakers. They may need to open the reference docs just to understand basic fields like `environment.type`, `engine.name`, `cases.files`, `expect`, and `judge`.

## Proposed solution

Add short field comments to the generated YAML examples and templates.

The comments should explain the meaning of important fields in a simple way. For example:

- `environment.type`: where the eval case runs
- `engine.name`: which agent engine is used
- `cases.files`: which case YAML files are included
- `expect`: cheap checks before judge runs
- `judge`: how the case result is graded

It would also be helpful if the `skill-upper` skill can generate these comments in English or Chinese depending on the user language. The YAML field names should stay in English.

## Alternatives considered

Only documenting these fields in reference docs is useful, but generated YAML files are usually the first thing users edit. Simple inline comments can make the first-time experience easier.

## Additional context

This should be a readability improvement only. It should not change the YAML schema or CLI behavior.

googs1025:
## Summary
- add field-leading comments to skill-upper eval and case YAML templates
- clarify generated YAML comment language rules while preserving technical identifiers
- extend skill-upper scaffold eval checks for English-only output and key field comments

## Verification
- ruby -e 'require "yaml"; YAML.load_file("skills/skill-upper/assets/eval.yaml.tmpl"); YAML.load_file("skills/skill-upper/assets/case.yaml.tmpl"); puts "PASS templates"'
- GOTOOLCHAIN=go1.26.4 make verify
- GOTOOLCHAIN=go1.26.4 GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=commit.gpgsign GIT_CONFIG_VALUE_0=false make test

Closes #174

### DIFF
--- a/skills/skill-upper/assets/eval.yaml.tmpl
+++ b/skills/skill-upper/assets/eval.yaml.tmpl
@@ -1,31 +1,45 @@
+# Evaluation config schema version used by skill-up. Do not change this value.
 schema_version: v1alpha1
 
 environment:
-  type: none                      # none | opensandbox
-  # kwargs:                        # For opensandbox: base_url, extensions, etc. See references/eval-yaml.md
+  # Runtime environment used for each case; use none when no sandbox is needed.
+  # Options: none | opensandbox | docker.
+  type: none
+  # Extra runtime settings for opensandbox or other selected environments.
+  # kwargs:
   #   base_url: https://agent-sandbox.example.com
+  # Container image used by sandboxed environments when a custom image is needed.
   # image: registry.example.com/your-org/sandbox-base:latest
+  # Workspace mount path inside the execution environment.
   # workspace_mount: /workspace
+  # Environment variables injected before setup steps and agent execution.
   # env:
   #   TZ: UTC
+  # Commands to prepare the workspace before each case starts.
   # setup_steps:
   #   - run: git config --global user.name "eval"
 
 skills:
+  # Skills installed into the Agent Engine before running cases.
+  # source options: local_path.
   - source: local_path
-    path: .                       # Directory of the Skill under test
+    # Path to the Skill under test, relative to this Skill root unless absolute.
+    path: .
   # - source: local_path
-  #   path: ../dependency_skill   # Path to other SKILL.md directory
+  #   path: ../dependency_skill
 
 engine:
-  name: claude_code               # claude_code | codex | qodercli (also compatible with qoder-cli)
-  # model is optional; omit to use engine's default model
+  # Agent Engine used to execute prompts and produce outputs for grading.
+  # Options: claude_code | codex | qodercli.
+  name: claude_code
+  # Optional model override; omit this block to use the engine default model.
   # model:
   #   provider: anthropic
   #   name: claude-sonnet-4-6
   #   base_url: ""
-  # kwargs:                         # agent-specific switches; see references/eval-yaml.md
-  #   bypass_sandbox: "true"        # codex: skip its own process sandbox (host kernel lacks Landlock)
+  # Agent-specific switches passed through to the selected engine.
+  # kwargs:
+  #   bypass_sandbox: "true"
 
 # mcp:
 #   servers:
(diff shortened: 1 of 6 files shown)


skills/skill-upper/assets/eval.yaml.tmpl
```diff
@@ -35,27 +49,41 @@ engine:
...
 #   model: anthropic/claude-sonnet-4-6
 #   context:
-#     profile: standard          # standard = file refs for transcript/diff; minimal = omit them
+#     profile: standard
 #   criteria:
 #     - "The response satisfies the case-specific success criteria."
 
 report:
+  # Report formats written after each run.
+  # Options: json | junit | html | grading | benchmark.
```
zpzjzj: **🟠 Major: Keep report format options aligned with the runner**

`report.formats` is only dispatched for `json`, `junit`, and `html` (`internal/runner/runner.go`). `grading.json` and benchmark artifacts are written independently, so listing `grading` and `benchmark` here teaches users unsupported/no-op format values.

**Suggestion**: list only `json | junit | html`; if useful, explain the automatically generated grading and benchmark artifacts separately.

skills/skill-upper/assets/eval.yaml.tmpl
```diff
@@ -1,31 +1,45 @@
...
+    path: .
   # - source: local_path
-  #   path: ../dependency_skill   # Path to other SKILL.md directory
+  #   path: ../dependency_skill
 
 engine:
-  name: claude_code               # claude_code | codex | qodercli (also compatible with qoder-cli)
-  # model is optional; omit to use engine's default model
+  # Agent Engine used to execute prompts and produce outputs for grading.
+  # Options: claude_code | codex | qodercli.
```
zpzjzj: **🟠 Major: Include the supported `qwen_code` engine**

The built-in engine registry also accepts `qwen_code` (plus its aliases), but this new `Options:` comment omits it. Because the wording appears exhaustive, users may incorrectly conclude that Qwen Code is unsupported.

**Suggestion**: at minimum add `qwen_code`, or label the list as built-ins and note separately that other names require an `engine.custom` block.

zpzjzj (commented):
Reviewed against the linked #174 scope at `3087221`.

The implementation direction is sound: it adds field-leading comments to both templates, makes comment language follow the user context, and extends the skill-upper eval coverage without changing the configuration schema or CLI behavior.

I left two Major inline comments where the new option descriptions do not match the current implementation:

1. `engine.name` omits the supported `qwen_code` engine.
2. `report.formats` lists `grading` and `benchmark`, although the runner only dispatches `json`, `junit`, and `html`; grading and benchmark artifacts are produced separately.

After those factual lists are corrected, the PR should satisfy #174 cleanly. `make verify` and `make test` pass locally, and the current GitHub checks are green.

skills/skill-upper/assets/eval.yaml.tmpl
googs1025: Fixed in eaaf806: report.formats now lists only json, junit, and html, with grading/benchmark artifacts documented as automatic outputs.

skills/skill-upper/assets/eval.yaml.tmpl
googs1025: Fixed in eaaf806: added qwen_code to the built-in engine options.

skills/skill-upper/assets/eval.yaml.tmpl
```diff
@@ -35,27 +49,42 @@ engine:
...
     #     - "TODO"
-  # parallelism: 4                # Stateless cases can run in parallel; or use skill-up run --parallelism N to override
+  # Maximum number of cases to run at the same time when cases are stateless.
+  # parallelism: 4
 
 # benchmark:
 #   enabled: true
 
+# Cheap pre-judge checks; when these fail, the configured judge is skipped.
+# expect:
```
zpzjzj: **Major: Move this `expect` example under `cases.defaults`**

Uncommenting this block creates a root-level `expect`, but `EvalConfig` has no such field. The loader uses permissive YAML unmarshalling, so `skill-up validate` still succeeds while the check is silently ignored; the advertised pre-judge gate never runs.

**Suggestion:** remove this duplicate block or nest it under `cases.defaults.expect` (case-specific checks belong in case YAML).

skills/skill-upper/assets/eval.yaml.tmpl
```diff
@@ -1,31 +1,45 @@
...
 environment:
-  type: none                      # none | opensandbox
-  # kwargs:                        # For opensandbox: base_url, extensions, etc. See references/eval-yaml.md
+  # Runtime environment used for each case; use none when no sandbox is needed.
+  # Options: none | opensandbox | docker.
+  type: none
+  # Extra runtime settings for opensandbox or other selected environments.
+  # kwargs:
   #   base_url: https://agent-sandbox.example.com
+  # Container image used by sandboxed environments when a custom image is needed.
```
zpzjzj: **Major: Clarify that Docker requires `environment.image`**

The options comment now advertises `docker`, but this image comment says an image is only needed when a custom image is wanted. Docker has no default-image fallback: selecting `docker` and following the template as written fails validation with `environment.image is required when environment.type is docker`.

**Suggestion:** say that `image` is required for Docker and optional/runtime-dependent for the other environments.

zpzjzj (commented):
Full-pass follow-up on the current head. The two previously reported option-list issues are fixed; the two remaining actionable findings are included together below.

skills/skill-upper/assets/eval.yaml.tmpl
googs1025: Fixed in e0e35f3: clarified that environment.image is required for docker and runtime-dependent for other environments.

skills/skill-upper/assets/eval.yaml.tmpl
googs1025: Fixed in e0e35f3: moved the must_contain example under cases.defaults.expect and removed the invalid root-level expect block.

zpzjzj (approved):
LGTM

PR #188, opened 2026-08-14
fix(agent): report QoderCLI token usage from JSON output

### LINKED ISSUE #187: [Bug]QoderCLI evaluations report zero token usage when session data lacks usage
## Summary

When skill-up runs an evaluation through QoderCLI, the generated QoderCLI session does not reliably record token usage. As a result, token statistics derived from the session are written as zero in the evaluation outputs.

## Actual behavior

- The QoderCLI session has missing or zero values for input and output token usage.
- `result.json` and `report.json` therefore contain `0` for `input_token` and `output_token` when skill-up relies on the session.
- In a real evaluation, QoderCLI's terminal JSON output contained non-zero `usage` values, while the session did not reliably contain the corresponding data.
- Setting `QODER_EXPOSE_TOKEN_USAGE=true` alone did not make session-based accounting reliable in the tested environment.

## Expected behavior

skill-up should report the actual QoderCLI token usage even when it is absent from the session. It can use QoderCLI's structured terminal JSON `usage` as a fallback and merge it with session-derived usage when available.

## Suggested acceptance criteria

- Successful QoderCLI evaluations produce non-zero input and output token counts in `result.json` and `report.json`.
- Cache token fields are preserved when QoderCLI provides them.
- Token accounting works for both with-skill and without-skill evaluation runs.
- Session usage remains the preferred richer source when available, with structured stdout used as a reliable fallback.
- Automated tests cover missing, zero, and populated session usage.

JHWang-1997:
## Summary

Report actual token usage for QoderCLI evaluations even when Qoder's private session files contain zero usage values or cannot be read from the evaluation runtime.

## Root cause

The QoderCLI adapter previously derived token counts only from the on-disk session JSONL. Qoder sessions commonly contain zero values for `input_tokens` and `output_tokens`, and isolated runtimes may not be able to access Qoder's private session directory at all. The adapter also captured plain-text stdout, so it had no structured fallback source for usage data.

## Changes

- Enable `QODER_EXPOSE_TOKEN_USAGE=true` by default while preserving an explicit user override.
- Request `--output-format json` for initial and resumed QoderCLI runs.
- Parse token usage and the session ID from the terminal JSON result.
- Keep an available session as the richer transcript source and merge its token high-water marks without replacing non-zero stdout usage.
- Persist QoderCLI terminal output as `stdout.json`.
- Update agent documentation, changelog, and unit coverage.

## Impact

`result.json` and `report.json` can now contain real QoderCLI input and output token counts when session-based accounting is unavailable or reports zeros. Existing session transcript behavior remains intact.

## Validation

- `make verify`
- `go test -race ./...`
- Manual QoderCLI evaluation confirmed non-zero token usage in generated reports.

Fixes #187


zpzjzj (commented):
I also validated this change against real QoderCLI releases and confirmed that the implementation works as intended:

- With QoderCLI `1.1.21`, the terminal JSON reports zero token values when `QODER_EXPOSE_TOKEN_USAGE` is unset, and non-zero usage when it is set to `true`.
- At this PR HEAD (`c0e56cf`), `TestAgent_QoderCLI_NoneRuntime_FullRun` passed with an isolated `HOME` and QoderCLI `1.1.21`. The generated `stdout.json` reported `input=18779`, `cache_read=16896`, and `output=28`; the final `result.json` reported `input_tokens=35675`, `output_tokens=28`, and `total_tokens=35703`.
- The QoderCLI version currently pinned in CI (`1.0.14`) also supports this environment variable, so the existing pin does not affect the correctness of this PR. The following items are therefore non-blocking recommendations.

Suggested follow-ups:

1. Update the QoderCLI pins in `.github/workflows/model-e2e.yml` and `action/Dockerfile` to the latest tested release, `1.1.21`, keeping the live E2E workflow and runner image aligned. The Linux x64 SHA-256 is `f979e907faa7ec5d74759c4cf87f2357d4f82eab35dadf98692a5a43c1055cdb`.
2. `TestAgent_QoderCLI_NoneRuntime_FullRun` currently only logs `result.json`. In the authenticated success path, add assertions that `input_tokens > 0` and `output_tokens > 0`, then run `Model E2E / none-runtime` manually. This branch currently has no corresponding live workflow run, while the regular PR checks cover only mocked JSON parsing.

The core fix has been validated end to end and looks ready to merge. The CI version update and stronger live assertions can be included as a separate commit in this PR or handled in a follow-up PR.

zpzjzj:
One additional documentation consistency issue: this PR correctly updates `internal/agent/README.md` to describe QoderCLI JSON output in `stdout.json`, but `internal/runtime/README.md` still lists `stdout.txt` as QoderCLI plain-text output.

Please update the runtime documentation in this PR as well. For example, the two engine-specific entries could be consolidated into:

```text
│   ├── stdout.json  # Claude Code JSONL / QoderCLI terminal result JSON
```

This will keep the artifact layout consistent with the implementation and avoid preserving the outdated assumption that QoderCLI stdout cannot provide token usage.

JHWang-1997:
Thanks @zpzjzj — addressed the review feedback in commit `a3c21b8`.

- Updated the QoderCLI pins in `.github/workflows/model-e2e.yml` and `action/Dockerfile` to `1.1.21`, keeping both locations aligned. I also independently verified the Linux x64 SHA-256 against Qoder's version-specific official manifest and the downloaded archive.
- Strengthened `TestAgent_QoderCLI_NoneRuntime_FullRun`: after a successful CLI run, the test now requires `result.json`, at least one case result, and non-zero aggregate `input_tokens` / `output_tokens`.
- Updated `internal/runtime/README.md` so the artifact layout consistently documents `stdout.json` for QoderCLI terminal result JSON.

Validation completed:
- `make verify`
- `make test`
- `make e2e`
- Authenticated local QoderCLI none-runtime E2E: `input_tokens=19608`, `output_tokens=8`

The local authenticated run used the installed QoderCLI `1.1.9`; the pinned `1.1.21` Linux artifact and checksum were verified independently. A GitHub `Model E2E / none-runtime` workflow dispatch has not been claimed here.

zpzjzj (approved):
LGTM

PR #191, opened 2026-08-14
fix(report): expose per-case execution metrics

### LINKED ISSUE #190: fix(report): distinguish evaluation wall time and expose per-case token usage
## Description

The generated evaluation report currently mixes different metric scopes:

- The top-level `Duration` is evaluation wall-clock time (`EndTime - StartTime`), including orchestration, tested-agent execution, judging, artifact handling, and framework overhead.
- Each case displays only the tested-agent session duration, but that scope is not identified clearly.
- Although `result.json` contains per-case `input_tokens` and `output_tokens`, the HTML report only shows one aggregate token total.

This makes it difficult to evaluate the actual time and token cost of running a Skill.

This is a generic reporting issue independent of the selected Agent engine. Agent-specific token acquisition problems are out of scope.

## Steps to reproduce

1. Run an evaluation in benchmark mode with at least two cases.
2. Open the generated HTML report.
3. Compare the header duration with per-case durations and inspect token usage.

Example metrics from one run:

| Case | Configuration | Agent time | Input tokens | Output tokens |
|---|---|---:|---:|---:|
| Basic expression | with_skill | 47.6s | 247,863 | 2,482 |
| Basic expression | without_skill | 14.9s | 59,113 | 525 |
| Operator precedence | with_skill | 26.6s | 238,466 | 983 |
| Operator precedence | without_skill | 21.5s | 59,282 | 810 |

## Expected behavior

- Label the top-level duration explicitly as **Evaluation Wall Time**.
- Treat per-case tested-agent execution time as the primary Skill cost metric.
- Display per-case agent execution time and input/output/total tokens.
- In benchmark mode, display with-Skill, without-Skill, and delta values.
- Report agent-judge duration and tokens separately when available.
- Display tested-agent, judge, and overall token totals.
- Keep JSON, HTML, and Markdown metric semantics consistent.
- Preserve compatibility with the existing `total_tokens` JSON field.

## Actual behavior

- The report header uses the ambiguous label `Duration` for evaluation wall time.
- Case views do not identify the duration scope.
- HTML reports omit per-case token usage and only display an aggregate total.

## Environment

- skill-up version: current `main`
- OS / arch: macOS / arm64

## Acceptance criteria

- [ ] Wall time and tested-agent execution time are clearly distinguished.
- [ ] Every case/configuration displays input, output, and total agent tokens.
- [ ] Benchmark cases display with-Skill, without-Skill, and delta.
- [ ] Agent-judge metrics are recorded separately when available.
- [ ] Aggregate totals separate tested-agent, judge, and overall tokens.
- [ ] JSON, HTML, and Markdown reports use consistent semantics.
- [ ] Tests cover single-run and benchmark metrics.
- [ ] No Agent-specific token parsing behavior changes in this issue.

JHWang-1997:
## Summary

- distinguish evaluation wall time from per-case tested-agent execution time
- expose per-case input, output, and total tested-agent token usage in JSON, HTML, and Markdown reports
- capture agent-judge duration and token usage separately and add tested-agent, judge, and overall token totals
- keep benchmark metrics compact beside each case heading, including with-Skill, without-Skill, and delta annotations
- preserve the existing `total_tokens` JSON field for compatibility

## Root cause

The top-level report duration comes from `EndTime - StartTime`, while each case duration comes from the tested-agent session. The HTML report labeled both scopes ambiguously and omitted the per-case token fields that were already present in `CaseResult`.

Agent-judge sessions also exposed duration and token usage internally, but those metrics were not copied into the report model before `JudgeSession` was excluded from JSON serialization.

This change is engine-independent. Agent-specific token acquisition, including QoderCLI session parsing, is intentionally out of scope.

## User impact

Reports now make Skill execution cost visible without allowing metrics to dominate the response and grading content:

- the header labels total duration as **Evaluation wall time**
- compact case-heading annotations show with-Skill, without-Skill, and delta metrics
- input/output token details remain available on hover
- agent-judge cost is reported separately
- Markdown and JSON reports use the same metric semantics

## Validation

- `make fmt`
- `make verify`
- `go test -race ./...`
- full local benchmark evaluation of `expression-calculator` using QoderCLI:
  - 2 cases × with/without Skill
  - 3 PASS, 1 baseline FAIL, 0 ERROR
  - evaluation wall time matched Agent + Judge execution time with approximately 0.2s framework overhead
- generated HTML embedded JavaScript syntax and compact-layout assertions

Closes #190

### DIFF
--- a/internal/runner/runner.go
+++ b/internal/runner/runner.go
@@ -541,10 +541,14 @@ func evalResultToCaseResult(res *evaluator.EvalResult) report.CaseResult {
 		Response:      responseContent(res),
 		TurnResults:   evalTurnResultsToReport(res.TurnResults),
 	}
-	if res.Grading != nil && res.Grading.JudgeSession != nil {
-		cr.JudgeDurationMs = res.Grading.JudgeSession.DurationMs
-		cr.JudgeInputTokens = res.Grading.JudgeSession.InputTokens
-		cr.JudgeOutputTokens = res.Grading.JudgeSession.OutputTokens
+	judgeSession := res.JudgeSession
+	if judgeSession == nil && res.Grading != nil {
+		judgeSession = res.Grading.JudgeSession
+	}
+	if judgeSession != nil {
+		cr.JudgeDurationMs = judgeSession.DurationMs
+		cr.JudgeInputTokens = judgeSession.InputTokens
+		cr.JudgeOutputTokens = judgeSession.OutputTokens
 	}
 	if res.Error != nil {
 		cr.Error = res.Error.Error()
--- a/internal/runner/runner.go
+++ b/internal/runner/runner.go
@@ -511,13 +511,16 @@ func buildReportInput(skillName string, grouped map[string]*caseResults, caseIDs
 			caseResult := evalResultToCaseResult(cr.withSkill)
 			input.CaseResults = append(input.CaseResults, caseResult)
 			input.TotalTokens += caseResult.InputTokens + caseResult.OutputTokens
+			input.JudgeTokens += caseResult.JudgeInputTokens + caseResult.JudgeOutputTokens
 		}
 		if cr.withoutSkill != nil {
 			caseResult := evalResultToCaseResult(cr.withoutSkill)
 			input.CaseResults = append(input.CaseResults, caseResult)
 			input.TotalTokens += caseResult.InputTokens + caseResult.OutputTokens
+			input.JudgeTokens += caseResult.JudgeInputTokens + caseResult.JudgeOutputTokens
 		}
 	}
+	input.OverallTokens = input.TotalTokens + input.JudgeTokens
 
 	return input
 }
@@ -537,6 +540,11 @@ func evalResultToCaseResult(res *evaluator.EvalResult) report.CaseResult {
 		Prompt:        res.Prompt,
 		Response:      responseContent(res),
 		TurnResults:   evalTurnResultsToReport(res.TurnResults),
+	}
+	if res.Grading != nil && res.Grading.JudgeSession != nil {
+		cr.JudgeDurationMs = res.Grading.JudgeSession.DurationMs
+		cr.JudgeInputTokens = res.Grading.JudgeSession.InputTokens
+		cr.JudgeOutputTokens = res.Grading.JudgeSession.OutputTokens
 	}
 	if res.Error != nil {
 		cr.Error = res.Error.Error()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- HTML and Markdown reports now identify the top-level duration as evaluation
+  wall time and show per-case tested-agent execution time plus input, output,
+  and total token usage. Benchmark cases include compact with-Skill,
+  without-Skill, and delta annotations beside the case heading. Agent-judge
+  time and tokens are reported separately and included in an explicit overall
+  token total.
 - Codex JSONL parsing now accepts records up to a configurable 16 MiB default
   while rejecting oversized records explicitly. The stdout stream is written to
   a bounded-download artifact instead of being buffered without limit in host
--- a/internal/evaluator/README.md
+++ b/internal/evaluator/README.md
@@ -79,7 +79,8 @@ type EvalResult struct {
     TurnsTotal    int                // Planned total number of turns
     InputTokens   int                // Number of input tokens
     OutputTokens  int                // Number of output tokens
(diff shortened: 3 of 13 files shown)


internal/runner/runner.go
```diff
@@ -538,6 +541,11 @@ func evalResultToCaseResult(res *evaluator.EvalResult) report.CaseResult {
 		Response:      responseContent(res),
 		TurnResults:   evalTurnResultsToReport(res.TurnResults),
 	}
+	if res.Grading != nil && res.Grading.JudgeSession != nil {
+		cr.JudgeDurationMs = res.Grading.JudgeSession.DurationMs
+		cr.JudgeInputTokens = res.Grading.JudgeSession.InputTokens
+		cr.JudgeOutputTokens = res.Grading.JudgeSession.OutputTokens
```
zpzjzj: **🟠 Major: Retain metrics from failed agent-judge sessions**

When an agent judge runs but its output cannot be parsed or validated, `runJudgePhaseWithSpan` recovers the available session from `SessionResultError` for artifact handling, but leaves `res.Grading` nil. This branch therefore drops the judge duration and token usage for ERROR cases, causing the new `judge_tokens` and `overall_tokens` totals to underreport actual execution cost.

**Suggestion:** preserve the recovered judge session on `EvalResult` and copy metrics from it here for both successful and failed judge executions. Please also extend the existing judge-error regression test with non-zero duration and token values and assert that they reach `CaseResult` and the aggregate totals.

---
*Source: Codex review @ `9a18eb6f`, with cross-file verification against `internal/evaluator/evaluator.go:650-660`.*

zpzjzj (commented):
## Code review summary

**Review scope:** `codex/fix-report-case-metrics` → `main` at `9a18eb6f`

### Findings

- Blocker: 0
- Major: 1
- Info: 1

### Compatibility assessment

The machine-readable `result.json` contract remains backward-compatible for normal JSON consumers:

- Existing fields such as `total_tokens`, `duration_ms`, `input_tokens`, and `output_tokens` are retained.
- `total_tokens` preserves its previous meaning: tested-agent tokens only.
- Judge and overall metrics are additive fields.
- Regenerating reports from the existing legacy fixture succeeds and preserves `total_tokens=12500`, while deriving `overall_tokens=12500`.

There is still a compatibility consideration for consumers that parse derived presentation formats. The HTML embedded `DATA` object renames `duration` / `total_tokens` and per-case `duration_ms` / `duration`, while the Markdown table schema also changes. No repository or publicly searchable downstream consumer was found, so this is informational rather than blocking. If these representations are consumed by internal automation, consider retaining the old HTML aliases for one release or explicitly documenting that only `result.json` is a stable machine-readable contract. Strict JSON schemas using `additionalProperties: false` should also be checked before rollout because the new additive fields would require a schema update.

### Overall assessment

⚠️ **Conditional merge.** The existing JSON fields are preserved, but the failed agent-judge path should retain its available session metrics before the new judge and overall totals are treated as reliable business cost data. The inline Major finding contains the concrete data-flow evidence and suggested regression coverage.

Validation reviewed: all GitHub checks are green; `make test` and targeted race-enabled report, runner, and CLI tests pass.

internal/runner/runner.go
JHWang-1997: Thanks for tracing this path. After revisiting the intended metric semantics, I don't think metrics from a failed agent-judge session should be included in these report aggregates.

This report is intended to describe valid evaluation samples, rather than serve as a billing or raw resource-consumption ledger:

- tested-agent metrics describe the Skill/case execution being measured;
- judge metrics are recorded only when the judge produces a parseable, validated grading result;
- if the judge fails because of an API/transport error, timeout, invalid JSON, or response-schema validation error, the case is marked `ERROR` and that judge run is an invalid measurement sample.

Including partial metrics from such a failed judge run in `judge_tokens` and `overall_tokens` would mix infrastructure/tooling anomalies into the statistics used to compare case execution. The recovered `SessionResult` should still be retained for artifacts and diagnostics, but not promoted into the evaluation aggregates.

Given that semantic choice, I propose to keep the current data flow, explicitly document that judge/overall totals include successfully validated judge sessions only, and add regression coverage confirming that failed judge-session metrics are excluded. If billing-grade accounting is needed later, it should be exposed as a separate raw usage/telemetry metric instead of overloading benchmark metrics.

JHWang-1997: Thanks — addressed in `97ca555`.

`EvalResult` now preserves the separate agent-judge `JudgeSession` independently of `Grading`, including the error path. The runner copies its duration and token usage into `CaseResult` and includes those tokens in `judge_tokens` / `overall_tokens`, while the case remains `ERROR` and `Grading` remains nil when no valid judgment was produced.

Added evaluator and runner regression coverage, updated the metric semantics docs, and verified with `make verify` and `make test`.

zpzjzj (approved):
LGTM

Issue #192, opened 2026-08-14
fix(judge): enforce a strict program-owned output contract for agent_judge


zpzjzj:
## Summary

`agent_judge` currently relies on the judge agent to return free-form text containing a JSON object:

```json
{
  "results": [
    {
      "criterion": "<criterion text>",
      "passed": true,
      "evidence": "..."
    }
  ]
}
```

The prompt describes this shape clearly, but the program does not enforce it as a strict output contract. In dogfooding with complex Skills and rubrics, judge output can be malformed or structurally valid but semantically misaligned, making evaluation results unstable.

The final `grading.json`, `result.json`, pass rate, status, and HTML report are already generated by skill-up. This issue is specifically about hardening the intermediate contract between the judge agent and skill-up.

## Background and validation priority

This issue was reported while using skill-up to evaluate complex data-analysis Skills, where the judge must follow detailed rubrics and return a stable machine-readable result.

The first validation target should be the current dogfooding combination:

- **Judge Agent engine:** Qoder (`qodercli`)
- **Judge model:** Qoder's **极致模型** tier/model

Implementation and regression verification should prioritize this combination before expanding the compatibility matrix to other Agent Engines and models. The exact resolved engine/model should be recorded in test evidence so model-specific behavior is not confused with framework-level contract enforcement.

The contract itself must remain engine-independent: native structured output may be used when supported, while strict text-JSON validation and bounded correction retry remain the compatibility path.

## Current behavior

The current implementation in `internal/judge/agent_judge.go`:

- extracts JSON from plain text, Markdown fences, or embedded objects;
- attempts to repair unescaped quotes;
- unmarshals with the permissive default `encoding/json` behavior;
- validates only that:
  - the number of returned results equals the number of configured criteria;
  - every result has non-empty `evidence`.

This leaves several gaps:

1. Unknown fields such as `name`, `score`, `threshold`, or `description` are silently ignored.
2. A missing `passed` field is indistinguishable from an explicit `passed: false` because the Go zero value is `false`.
3. A missing, rewritten, duplicated, reordered, or mismatched `criterion` is accepted as long as the result count and evidence pass validation.
4. The report uses the agent-returned criterion text instead of the configured criterion as the source of truth.
5. Evidence is only checked for non-emptiness; any stronger evidence contract remains prompt-only.
6. Invalid output immediately becomes a judge `ERROR`; there is no bounded correction retry.

## Proposed direction

Make the judge result contract program-owned rather than prompt-owned.

[...]

PR #197, opened 2026-08-17
fix(judge): enforce strict agent judge output contract

DancingCircles:
## Summary

- assign stable criterion IDs and keep configured criterion text authoritative
- enforce a strict JSON response contract with required decision, evidence, and failure fields
- deterministically map judge results and preserve invalid judge sessions as errors
- add regression coverage and a credential-gated QoderCLI ultimate-model E2E test

## Related issue

Refs #192

This is PR1 of the maintainer-approved split. The bounded correction retry and preservation of both raw outputs remain for PR2.

## Test plan

- `make verify`
- `make test` (`go test -race ./...`)
- `go test -tags e2e -timeout 1200s -count=1 ./e2e`
- verified the QoderCLI ultimate-model E2E skips when full-E2E credentials are unavailable; the existing trusted Model E2E workflow runs it when credentials are present


internal/judge/agent_judge.go
```diff
@@ -210,194 +213,142 @@ func buildContextMetadata(sessionResult *agent.SessionResult, materialized *Mate
...
+	lineEnd := strings.IndexByte(trimmed, '\n')
+	if lineEnd < 0 || !strings.EqualFold(strings.TrimSpace(trimmed[3:lineEnd]), "json") {
+		return "", errors.New("agent_judge response must be JSON or a single JSON code fence")
+	}
+	remainder := strings.TrimSpace(trimmed[lineEnd+1:])
+	if !strings.HasSuffix(remainder, "```") {
+		return "", errors.New("agent_judge JSON code fence is not closed")
 	}
+	candidate := strings.TrimSpace(strings.TrimSuffix(remainder, "```"))
+	if strings.Contains(candidate, "```") {
```
zpzjzj: **[P2] Allow fence markers inside JSON string values**

After the outer closing fence is removed, this substring check rejects any remaining triple-backtick sequence, including one inside a valid JSON string. An `evidence` or `failures` entry may legitimately quote a Markdown code block, so a fully valid fenced contract response would become a judge `ERROR` here.

The strict JSON decoder below already rejects real extra fences, trailing content, and additional JSON values. Please remove this blanket check or identify fence boundaries structurally, and add a regression case where fenced JSON contains a code-fence marker inside an evidence string.

zpzjzj (commented):
Overall, this PR is well aligned with the PR1 scope described in #192. It implements stable criterion IDs, keeps configured criterion text authoritative, rejects missing/unknown contract fields, validates and deterministically reorders criterion results, preserves an invalid judge session through `SessionResultError`, and adds the credential-gated QoderCLI `ultimate` E2E coverage requested by the issue.

The two intentionally deferred items—one bounded validation-guided correction retry and preservation of both raw judge outputs—belong to PR2, so their absence here is consistent with the stated split. Native structured-output capabilities can also remain incremental as proposed by the issue.

I found one P2 correctness issue in the fenced-JSON parser, noted inline. Once that is addressed, this PR1 looks consistent with the issue's correction plan.

Validation performed at `5face8b`:
- `make verify` passed
- `go test -race ./internal/judge ./internal/evaluator ./internal/cli` passed
- the QoderCLI Ultimate E2E compiled and skipped as expected without full-E2E credentials
- GitHub currently reports only the passing CLA check; no Actions result is visible yet

Review filtering: 1 finding accepted, 0 downgraded, 0 rejected. The bundled Claude Code `/code-review` entrypoint was unavailable locally, so the cold-read source was `codex exec review`, followed by cross-file and test evidence checks.

internal/judge/agent_judge.go
DancingCircles: Fixed in e055557. I removed the blanket triple-backtick substring check and added a regression case proving that a fenced JSON response may contain a code-fence marker inside an evidence string. The strict decoder continues to reject actual trailing fences and extra JSON values. Validation: `make verify` and `go test -race ./...` both pass.

zpzjzj (commented):
Re-review of e055557:

The previous fenced-JSON issue is resolved correctly: removing the blanket triple-backtick check allows fence markers inside JSON strings, while the strict decoder still rejects trailing content and multiple JSON values. The implementation remains aligned with the PR 1 scope in #192: a stricter structured contract, stable criterion IDs, deterministic mapping, and program-side validation/aggregation. Correction retry and preservation of both raw attempts remain appropriately deferred to PR 2.

One remaining issue:

**[P2] Update the shipped Judge debug fixture for the new contract**

`examples/judge-debug-agent.json` still uses the old `criterion` / `reason` mock-result shape. `judgeDebugInput.MockResults` is unmarshaled directly into `[]judge.CriterionResult`, so running `skill-up debug judge examples/judge-debug-agent.json` now fails with `agent_judge: result 1 is missing criterion_id`. Please migrate the example to `criterion_id`, `evidence`, and `failures`, and preferably add an executable regression test covering the shipped fixture. The regular test suite currently does not catch this.

Validation: `make verify`, `go test -race ./internal/judge ./internal/cli`, and the full `make test` run pass; the shipped debug command reproduces the fixture failure above.

Review filtering: 1 accepted finding, 0 downgraded, 0 rejected; the prior finding is resolved and not repeated.

DancingCircles:
Addressed in 028c7f8. I migrated `examples/judge-debug-agent.json` to the strict `criterion_id` / `evidence` / `failures` contract and added an executable CLI regression test that reads and runs the shipped fixture directly.

Validation:
- `go run ./cmd/skill-up debug judge examples/judge-debug-agent.json --output /tmp/grading.json` now completes and reports the expected `FAIL — 2/3 assertions passed` at the fixture's `0.7` threshold
- `go test -race ./internal/cli ./internal/judge` passes
- `make verify` passes with 0 issues
- `go test -race ./...` passes

Thanks for catching the stale fixture.

internal/judge/agent_judge.go
```diff
@@ -210,194 +213,139 @@ func buildContextMetadata(sessionResult *agent.SessionResult, materialized *Mate
...
-func findJSONObjectCandidates(output string) []string {
-	candidates := make([]string, 0)
-	start := strings.Index(output, "{")
-	for start >= 0 {
-		end, ok := findJSONObjectEnd(output, start)
-		if !ok {
-			break
-		}
-		candidates = append(candidates, output[start:end+1])
+	decoder := json.NewDecoder(strings.NewReader(candidate))
```
zpzjzj, line 315: **[P2] Reject duplicate JSON object keys**

`DisallowUnknownFields` rejects unknown members, but Go's JSON decoder still accepts duplicate known members with last-value-wins semantics. For example, a judge response containing both `"passed": false` and `"passed": true` is accepted, and the semantic validator can turn that ambiguous malformed response into PASS instead of ERROR. That undermines the strict program-owned contract introduced by this PR.

Please detect duplicate member names before decoding into `judgeResponse` (including nested result objects) and add a regression test for duplicate `passed` or `criterion_id` keys.

e2e/agent_test.go
```diff
@@ -1485,6 +1485,107 @@ report:
...
+			Evidence string `json:"evidence"`
+		} `json:"expectations"`
+		Summary struct {
+			PassRate float64 `json:"pass_rate"`
+		} `json:"summary"`
+	}
+	if err := json.Unmarshal(data, &grading); err != nil {
+		t.Fatalf("parse grading.json: %v", err)
+	}
+	if grading.Summary.PassRate != 1 || len(grading.Expectations) != 1 || !grading.Expectations[0].Passed {
```
zpzjzj, line 1575: **[P2] Fix the Qoder ultimate E2E grading assertions**

This case configures both `expect.must_contain` and `agent_judge`. After a successful expect pre-check, `evaluateCaseSession` prepends the expect assertion to `finalResult.Grading.AssertionResults`, and `grading.json` serializes every assertion. The credentialed run therefore produces two expectations, with `expect.must_contain` first; this check requires exactly one and the following assertion treats index 0 as the configured criterion, so the new trusted Qoder ultimate E2E will fail even when both checks pass.

Please either remove the redundant expect block or assert both entries and locate the agent-judge criterion by its configured text instead of assuming index 0.

zpzjzj (commented):
Consolidated review of the current head. The two earlier findings are resolved; these are the two remaining actionable issues found in the full pass. Both are included in this single review to avoid serial review/fix cycles.

Validation: normal GitHub CI is green; `make verify` and targeted race tests passed. The credential-gated Qoder ultimate E2E issue below is not exercised by the normal PR checks.

internal/judge/agent_judge.go
DancingCircles: Fixed in b9d6bad. The decoder now performs a recursive token-stream scan before struct decoding and rejects duplicate keys in every JSON object, including the root response and nested criterion results. Regression cases cover duplicate `results`, `criterion_id`, and conflicting `passed` fields. `DisallowUnknownFields` and trailing-value checks remain in place.

e2e/agent_test.go
DancingCircles: Fixed in b9d6bad. I removed the redundant `expect.must_contain` block, so this credential-gated case now produces exactly one grading expectation from `agent_judge`; the existing text and evidence assertions therefore target the intended criterion. The focused E2E compiles and skips as expected without full credentials, and the full quick E2E suite passes.

zpzjzj (commented):
Consolidated re-review of the current head: 0 Blocker, 1 Major, 0 Info. All 11 changed files and the strict-decoding contract, callers, report serialization, tests, fixtures, and gated E2E path were checked. The two previous findings are resolved. Local `make verify` and `make test` pass; GitHub CI is green. The credential-gated Qoder Ultimate run could only be compiled and confirmed to skip without `QODER_PERSONAL_ACCESS_TOKEN`. Filtering: 1 finding accepted, 0 downgraded, 0 rejected. This review contains the only remaining actionable issue found in the full pass.

internal/judge/agent_judge.go
```diff
@@ -210,194 +213,194 @@ func buildContextMetadata(sessionResult *agent.SessionResult, materialized *Mate
...
 			}
-			if inDoubleQuotedString && ch == '"' {
-				inDoubleQuotedString = false
+			key, ok := keyToken.(string)
+			if !ok {
+				return errors.New("JSON object key is not a string")
 			}
-			if inSingleQuotedString && ch == '\'' {
-				inSingleQuotedString = false
+			if _, exists := seen[key]; exists {
```
zpzjzj, line 372: **🟠 Major: Reject case-variant aliases before struct decoding**

The duplicate-key scan treats `passed` and `Passed` as different keys, but Go's `encoding/json` matches struct fields case-insensitively. A response containing `"passed": false` followed by `"Passed": true` is therefore accepted and the later value overwrites the earlier one, allowing ambiguous judge output to flip the grading result despite the strict contract. The same applies to aliases such as `results` / `Results`.

Please validate exact allowed key names at each contract level before struct decoding (or otherwise reject case-variant aliases), and add a regression test covering a conflicting `passed` / `Passed` pair.

internal/judge/agent_judge.go
```diff
@@ -210,194 +213,194 @@ func buildContextMetadata(sessionResult *agent.SessionResult, materialized *Mate
...
-	candidates := make([]string, 0)
-	start := strings.Index(output, "{")
-	for start >= 0 {
-		end, ok := findJSONObjectEnd(output, start)
-		if !ok {
-			break
-		}
-		candidates = append(candidates, output[start:end+1])
+	decoder := json.NewDecoder(strings.NewReader(candidate))
+	decoder.DisallowUnknownFields()
```
JHWang-1997, line 316: **Suggestion (non-blocking): consider enforcing exact field-name casing**

One small strictness edge case remains here: Go's `encoding/json` matches struct fields case-insensitively even when `DisallowUnknownFields` is enabled. As a result, a response using keys such as `RESULTS`, `CRITERION_ID`, or `PASSED` is still accepted as the advertised lowercase contract and can produce a valid result.

Since this PR establishes a program-owned strict schema, it may be worth validating the exact allowed key spellings during the token scan that now detects duplicate keys. That would keep the checks in one place; a small regression case with case-folded keys would document the intended behavior.

JHWang-1997 (commented):
One small, non-blocking suggestion on the latest strict-decoding follow-up.

internal/judge/agent_judge.go
DancingCircles: Fixed in 2b2d178. The pre-decode token scan is now contract-level aware: the root object accepts only the exact `results` spelling, and each result object accepts only exact `criterion_id`, `passed`, `evidence`, and `failures` spellings. This rejects conflicting `passed` / `Passed` aliases before `encoding/json` can overwrite the value. Regression cases cover `RESULTS`, `CRITERION_ID`, and the conflicting pair. `make verify`, `go test -race ./...`, and the quick E2E suite pass.

internal/judge/agent_judge.go
DancingCircles: Thanks — addressed in 2b2d178 as part of the strict key scan. Exact casing is now enforced at both contract object levels, with regression coverage for uppercase root/result keys and mixed-case conflicting aliases.

zpzjzj (approved):
LGTM

PR #198, opened 2026-08-17
fix(runner): scope report timing to each iteration

### LINKED ISSUE #195: Multi-iteration reports use cumulative wall time and reuse the same start_time
## Description

When `skill-up run ... --iteration N` performs multiple samples, each `iteration-N/` directory contains only that iteration's results, but its report timestamps do not use iteration-local boundaries.

`Runner.startTime` is set once before the iteration loop, while `Runner.endTime` is updated after every iteration. As a result, iteration N reports the cumulative wall time from the beginning of iteration 1 through the end of iteration N. Every iteration also gets the same `start_time`.

This is present on current `main` at commit `3abde0864ceae91c1ae000c516d15ba6534c7471`.

## Steps to reproduce

1. Run an evaluation with three iterations and report formats enabled:

   ```bash
   skill-up run ./evals/eval.yaml --iteration 3 --format html --format junit
   ```

2. Assume each iteration takes approximately 100 seconds.
3. Inspect `result.json`, `report.html`, and `report.xml` under each `iteration-N/` directory.

## Expected behavior

Each report should describe only its own iteration:

| Directory | start_time | end_time | Wall time |
|---|---|---|---:|
| `iteration-1/` | T0 | T0+100s | 100s |
| `iteration-2/` | T0+100s | T0+200s | 100s |
| `iteration-3/` | T0+200s | T0+300s | 100s |

The exact timestamps can include per-iteration orchestration overhead, but each interval should be local to that iteration.

## Actual behavior

| Directory | start_time | end_time | Reported wall time |
|---|---|---|---:|
| `iteration-1/` | T0 | T0+100s | 100s |
| `iteration-2/` | T0 | T0+200s | 200s |
| `iteration-3/` | T0 | T0+300s | 300s |

In practice, later durations may be even higher because they also include report generation and workspace setup between earlier iterations.

A single iteration is unaffected because the command-level and iteration-level time boundaries happen to coincide.

## Impact

The incorrect timestamps are persisted in `result.json` and affect all consumers of `report.Input.TotalDuration()` / `StartTime`:

- HTML: Start Time and Evaluation Wall Time
- JUnit: `<testsuite time="..." timestamp="...">`
- Markdown generated from `result.json`: Evaluation Wall Time
- JSON reports regenerated from the saved input

The JUnit values may pollute CI duration statistics and trend charts.

## Root cause

In `internal/runner/runner.go`:

- `r.startTime = time.Now()` is outside the iteration loop.
- `r.endTime = time.Now()` is inside the loop.
- `WriteResults` receives only the current iteration's `results`, but `buildReportInput` uses those shared Runner timestamps.
- `report.Input.TotalDuration()` calculates `EndTime - StartTime` and is reused by the HTML, JUnit, and Markdown reporters.

## Suggested fix

Capture a fresh start and end time for every iteration and bind those values to that iteration's report input. Prefer iteration-local variables such as `iterationStartTime` and `iterationEndTime` over Runner-level mutable timing fields.

## Acceptance criteria

- Each `iteration-N/result.json` contains a time range covering only that iteration.
- Later iterations have their own increasing `start_time` values.
- HTML, JUnit, and Markdown wall times no longer increase cumulatively.
- Single-run and `--iteration 0` auto-append behavior remain unchanged.
- A regression test covers timing boundaries across multiple iterations.

## Environment

- skill-up revision: `3abde0864ceae91c1ae000c516d15ba6534c7471` (`main`)
- OS / arch: independent of platform

JHWang-1997:
## Summary

- capture a fresh start and end time for every evaluation iteration
- pass iteration-local timing explicitly into `WriteResults` instead of storing it on the shared `Runner`
- add deterministic regression coverage that verifies independent 100-second timing windows across multiple iterations
- document the fix in the changelog

## Root cause

`Runner.startTime` was initialized once before the iteration loop, while `Runner.endTime` was updated after each iteration. Every per-iteration report therefore reused the command start time and reported a cumulative duration.

Because `result.json` is the source for the HTML, JUnit, JSON, and Markdown reporters, the incorrect timing propagated to all report formats. In particular, JUnit `testsuite.time` and `timestamp` could pollute CI duration statistics.

The timing values now have the same lifecycle as the iteration results they describe.

## Validation

- `make verify`
- `env -u QODER_PERSONAL_ACCESS_TOKEN make test`
- `env -u QODER_PERSONAL_ACCESS_TOKEN make e2e`
- `git diff --check`

Fixes #195

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   without-Skill, and delta annotations beside the case heading. Agent-judge
   time and tokens are reported separately and included in an explicit overall
   token total.
+- Multi-iteration reports now record each iteration's own start time and wall
+  time instead of reusing the command start time and accumulating the duration
+  of all preceding iterations.
 - Codex JSONL parsing now accepts records up to a configurable 16 MiB default
   while rejecting oversized records explicitly. The stdout stream is written to
   a bounded-download artifact instead of being buffered without limit in host
--- a/internal/runner/runner.go
+++ b/internal/runner/runner.go
@@ -55,8 +55,7 @@ type Runner struct {
 	runnerParams credential.AgentInitParams
 
 	workspace *report.IterationWorkspace
-	startTime time.Time
-	endTime   time.Time
+	now       func() time.Time
 }
 
 // NewRunner creates a new Runner with the given eval config and loader.
@@ -66,6 +65,7 @@ func NewRunner(evalCfg *config.EvalConfig, loader *config.Loader, resolver *cred
 		loader:       loader,
 		resolver:     resolver,
 		runnerParams: runnerParams,
+		now:          time.Now,
 	}
 }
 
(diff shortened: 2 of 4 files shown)


zpzjzj (approved):
LGTM

PR #209, opened 2026-08-18
fix(judge): add bounded correction retry for agent judge

### LINKED ISSUE #192: fix(judge): enforce a strict program-owned output contract for agent_judge

DancingCircles:
## Summary

- add exactly one validation-guided correction retry for strict `agent_judge` decode or semantic validation failures
- resume the Judge session when supported, with one serialized first-turn fallback message otherwise
- preserve every raw Judge response and isolate retry engine artifacts under `outputs/judge/run/retry/`
- aggregate both attempts into the final Judge session without changing report schemas
- add Judge, Evaluator, deterministic E2E, and credential-gated QoderCLI coverage

## Relationship to #197

This builds on #197, which introduced the strict program-owned response contract and deterministic criterion mapping. This PR completes the maintainer-approved PR2 scope: one bounded correction retry and raw-output artifact preservation.

Closes #192

## Behavior

- retry only after strict JSON decoding or semantic contract validation fails
- do not retry a valid FAIL judgment, explicit cancellation, context materialization failure, or an agent error without recoverable output
- share the existing Judge/case timeout budget across both attempts; no timeout reset or backoff
- use `agent.SessionResumer.RunTurn` when a session ID is available; otherwise call `Run` with one serialized user message containing the original prompt, first raw response, and correction instructions
- keep the last attempt authoritative while aggregating duration, token/turn usage, and stable-deduplicated artifacts
- keep native structured output, configurable retry policies, and report schema/UI changes out of scope

## Artifact layout

```text
outputs/judge/run/
├── stdout.json (or other first-attempt engine artifacts)
├── raw-response-attempt-1.txt
├── raw-response-attempt-2.txt       # only when correction runs
└── retry/
    └── stdout.json (or other retry engine artifacts)
```

## Test plan

- `go test -race ./internal/judge ./internal/evaluator ./internal/cli`
- `make fmt`
- `make verify`
- `go test -tags e2e -timeout 1200s -count=1 ./e2e`
- `go run ./cmd/skill-up debug judge examples/judge-debug-agent.json --output /tmp/grading.json`
- verified the deterministic mock performs exactly `initial → correction`, returns PASS, and preserves both raw responses plus retry artifacts
- verified the QoderCLI `ultimate` test skips without full-E2E prerequisites; the trusted Model E2E runner will execute it when QoderCLI and credentials are available

## Local Docker note

`go test -race ./...` passes for every package except the unchanged upstream test `internal/runtime.TestNoneRuntime_ExecStaleLockDoesNotHangNextCommand`, which reproducibly reports that a background child survives timeout in this Docker environment. `internal/runtime` has no diff from `upstream/main`; all other packages pass together under `-race`.


internal/judge/agent_judge.go
```diff
@@ -138,44 +148,254 @@ func (j *AgentJudge) Evaluate(ctx context.Context, in Input) (*Result, error) {
...
+		)
+	}
+
+	messages := make([]transcript.Message, 0, len(originalMessages)+2)
+	messages = append(messages, originalMessages...)
+	if firstSession != nil {
+		messages = append(messages, transcript.Message{Role: transcript.RoleAssistant, Content: firstSession.FinalMessage, Turn: 1})
+	}
+	messages = append(messages, transcript.Message{Role: transcript.RoleUser, Content: correctionPrompt, Turn: 2})
+	return j.Agent.Run(ctx, j.Runtime, opts, messages)
```
zpzjzj: **🟠 Major: Preserve the invalid response in the fallback correction prompt**

When the agent does not support `SessionResumer`, or the first result has no session ID, this branch passes the invalid response as an assistant message to a fresh `Agent.Run`. However, Qwen and the fresh-run paths of Qoder/Claude/Codex flatten messages through `BuildInstructionFromMessages`, which discards assistant messages. A custom engine using `${prompt}` receives an empty prompt because this artificial retry is now multi-message. The correction attempt therefore may never see `firstSession.FinalMessage`. Serialize the original prompt, invalid response, and correction instruction explicitly into one fallback user message, and cover the rendered adapter input in tests.

internal/judge/agent_judge.go
```diff
@@ -138,44 +148,254 @@ func (j *AgentJudge) Evaluate(ctx context.Context, in Input) (*Result, error) {
...
+func judgeSessionMetricsAreCumulative(first, second *agent.SessionResult) bool {
+	if len(first.Transcript) == 0 || len(second.Transcript) < len(first.Transcript) {
+		return false
+	}
+	for i := range first.Transcript {
+		if !reflect.DeepEqual(first.Transcript[i], second.Transcript[i]) {
+			return false
 		}
 	}
+	return true
```
zpzjzj, line 405: **🟠 Major: Do not infer cumulative usage from transcript prefixes**

A non-resumable custom `session_result` engine can omit its transcript, causing `minimalCustomTranscript` to build the second independent invocation from the replayed message history. The second transcript then starts with the first transcript, even though `input_tokens` and `output_tokens` still describe only that invocation. This check classifies those counters as cumulative and keeps only the maximum, so reports omit the first Judge attempt usage. Track whether the correction used an independent `Run` or a resumed `RunTurn`, and sum usage for independent retries instead of relying only on transcript prefixes.

zpzjzj (commented):
Review summary: 0 Blockers, 2 Major findings, and 1 Info item. I recommend addressing the Major findings before merging.

Coverage: reviewed all 10/10 changed files, including the correction contract, agent adapters and callers, evaluator/report metrics, artifact handling, tests, fixtures, documentation, E2E paths, and CI-only coverage. All current CI checks are green.

Info: please use an English commit subject. Replace `fix(judge): 增加有界纠错重试` with `fix(judge): add bounded correction retry`.

internal/judge/agent_judge.go
DancingCircles: Fixed in 422bdaf. The non-resumable path now sends one serialized user message containing the original Judge prompt, the JSON-escaped raw invalid response, and the correction instructions. This keeps the complete context visible to adapters that flatten only user messages and to custom `${prompt}` engines. Regression coverage now asserts the rendered `BuildInstructionFromMessages` input and runs a real CustomAgent `${prompt}` template; the deterministic correction E2E also passes with the new envelope.

internal/judge/agent_judge.go
DancingCircles: Fixed in 422bdaf. The correction call now returns an explicit independent-Run vs resumed-RunTurn mode. Independent retries always sum duration, turns, and token counters, even when their synthesized transcript has the first transcript as a prefix. Only resumed turns may use prefix detection to treat counters as cumulative. Added a regression case for an independent prefix-shaped transcript.

DancingCircles:
Thanks for the thorough review. Both Major findings are addressed in 422bdaf, with focused adapter/aggregation regressions plus deterministic and full quick E2E coverage. The follow-up commit and PR title use English subjects. I kept the already-reviewed first commit intact rather than force-pushing rewritten history; the final PR title is `fix(judge): add bounded correction retry for agent judge`.

internal/judge/agent_judge.go
```diff
@@ -138,44 +155,259 @@ func (j *AgentJudge) Evaluate(ctx context.Context, in Input) (*Result, error) {
...
+
+	invalidResponse := ""
+	if firstSession != nil {
+		invalidResponse = firstSession.FinalMessage
+	}
+	fallbackPrompt := buildAgentJudgeFallbackCorrectionPrompt(originalPrompt, invalidResponse, correctionPrompt)
+	sessionResult, err := j.Agent.Run(ctx, j.Runtime, opts, []transcript.Message{{
+		Role:    transcript.RoleUser,
+		Content: fallbackPrompt,
+		Turn:    2,
```
zpzjzj: **🟠 Major: Count an independent retry as a new first turn**

This branch invokes a fresh `Agent.Run`, but marks its only input as `Turn: 2`. A Custom Engine using the documented `session_result` contract may omit both `turns` and `transcript`; `parseSessionResult` then builds a minimal transcript and `deriveTurns` returns the maximum explicit turn. The independent retry consequently reports two turns, and `aggregateAgentJudgeSessions` sums the first and second runs so two actual invocations become three reported turns.

Set this independent-run message to `Turn: 1`; reserve turn 2 for the resumed `RunTurn` branch. Please add a real CustomAgent regression where the engine omits `turns` and `transcript`.

internal/judge/agent_judge.go
```diff
@@ -138,44 +155,259 @@ func (j *AgentJudge) Evaluate(ctx context.Context, in Input) (*Result, error) {
...
+	logging.WarnContextf(ctx, "agent_judge output failed validation; retrying once with correction guidance: %v", validationErr)
 
-	var resp judgeResponse
-	if err := decodeAgentJudgeResponse(sessionResult.FinalMessage, &resp); err != nil {
+	correctionPrompt := buildAgentJudgeCorrectionPrompt(j.Criteria, validationErr)
+	retryArtifactDir := ""
+	if in.ArtifactDir != "" {
+		retryArtifactDir = filepath.Join(in.ArtifactDir, agentJudgeRetryArtifactDir)
+	}
+	retrySession, correctionMode, retryErr := j.runAgentJudgeCorrection(ctx, firstSession, prompt, correctionPrompt, retryArtifactDir)
```
zpzjzj, line 190: **🟠 Major: Snapshot Custom Engine framework artifacts before the retry overwrites them**

Changing `ExecOptions.ArtifactDir` does not isolate the Custom Engine framework input/output. `buildBaseVars` and `resolveCustomIOFiles` still resolve both attempts to the same workspace paths, and `runLocal` overwrites those paths on the retry. The merged session then deduplicates the identical paths, while the evaluator copies them only after both attempts have completed. As a result, the first-attempt framework input/output are lost and the retry directory does not contain its own framework files, contrary to the artifact-isolation contract described by this PR.

Snapshot each attempt framework input/output into its `ArtifactDir` before another invocation can overwrite them, or allocate distinct runtime I/O paths per attempt. Add an evaluator-level CustomAgent correction test that verifies the first and retry contents separately.

zpzjzj (commented):
Review summary: 0 Blockers, 2 Major findings, and 2 Info items. I recommend addressing the Major findings before merging.

Coverage: reviewed all 11/11 changed files at `422bdaf`, including the correction contract, all built-in and custom agent paths, metric aggregation, artifact collection and archival, evaluator/report consumers, unit tests, fixtures, documentation, deterministic E2E, credential-gated paths, and current CI. `make test`, `make verify`, and the targeted correction E2E pass; all GitHub checks are green. The local full E2E suite reached and passed the new correction test, but the pre-existing QoderCLI live test failed because the successful external run reported zero token usage.

Info:
- Commit `7c9e1b1` still has a Chinese subject and body. Adding a later English commit and changing the PR title does not make the historical commit information English; please reword that commit.
- The Summary still says "deterministic three-message fallback", while the implementation and Behavior section now correctly describe one serialized user message. Please update the stale Summary text.

The two Major findings from the previous review have been fixed in `422bdaf`; the comments below are distinct issues found during the complete re-review.

internal/judge/agent_judge.go
DancingCircles: Fixed in aed0cdd. Immediately after each Judge call, runtime-backed GeneratedFiles are snapshotted into that attempt ArtifactDir before validation can launch another invocation; successful snapshots replace the overwrite-prone workspace paths in the session, while failures remain best-effort and preserve the original path. The evaluator-level real CustomAgent regression verifies distinct first/retry messages.json and session-result.json contents under judge/run and judge/run/retry.

internal/judge/agent_judge.go
DancingCircles: Fixed in aed0cdd. The non-resumable correction is a fresh Agent.Run and its serialized fallback message now uses Turn: 1; Turn: 2 remains reserved for RunTurn session resumption. The new evaluator-level real CustomAgent regression omits both turns and transcript from each SessionResult and asserts that two independent Judge invocations aggregate to exactly two turns.

DancingCircles:
Both new Major findings are addressed in `aed0cdd`, with a real evaluator-level CustomAgent regression covering fresh-turn accounting and separate framework input/output snapshots. The stale Summary wording has also been updated. Local verification: focused race tests pass, `make verify` reports 0 issues, all packages except the documented unchanged Docker-only `internal/runtime` case pass under race, and the full quick E2E suite passes.

For the remaining historical commit Info item: rewording `7c9e1b1` requires a non-fast-forward force push over the already-reviewed branch. I have not rewritten that remote history. If English-only individual commit history is mandatory, would you prefer a replacement clean PR branch; otherwise an English squash-merge commit using the current PR title preserves an English final history?

zpzjzj (approved):
LGTM

Issue #196, opened 2026-08-17
refactor(agent): clarify model, provider, protocol, auth, and applied config semantics


zpzjzj:
## Background

Agent configuration currently mixes several independent concepts:

- the agent CLI (`engine`);
- the adapter wire protocol;
- the opaque upstream model identifier;
- the provider namespace used for credentials and endpoints;
- injected authentication versus delegated agent-local login;
- values requested by the user, values applied by skill-up, and values actually observed from the running agent.

The problem is not the name or YAML nesting of `provider`. The problem is that provider, protocol, routing, credentials, and runtime observation are resolved in multiple layers and can be silently rewritten or reported as if they were equivalent.

Local login remains a supported workflow. An already logged-in Codex CLI, QoderCLI, Claude Code, or Qwen Code must be usable without injected auth. Preflight checks must remain static and must not make a model request or consume tokens.

## Intended semantics

```text
engine                 selects the CLI adapter and its default protocol
model.name             is an opaque upstream model identifier
model.provider         selects a credential/endpoint namespace
auth mode              is injected or delegated to agent-local state
requested config       is selected from YAML/CLI/env/credential inputs
applied config         is what skill-up forwards after adapter normalization
observed config        is only what the running agent explicitly reports
```

The engine, not the provider, determines the protocol. For example, `provider: dashscope` with Codex selects an OpenAI-compatible connection, while the same provider with Claude Code selects an Anthropic-compatible connection.

Missing observed values are unknown. Applied configuration must never be presented as proof of the provider, model, credential, or local login ultimately used by the agent.

## Current problems

1. `provider` is overloaded as a legacy model prefix, credential namespace, apparent vendor, and adapter input.
2. `--model X/Y` is ambiguous between a legacy `provider/model` form and an opaque slashed model ID.
3. Explicit configuration can be silently overridden or ignored (`<PROVIDER>_MODEL`, unsupported Qoder modes, Codex custom-provider fallback, ineffective kwargs).
4. Credential files can describe protocol-specific endpoints, but current resolution flattens them before the adapter protocol is known.
5. Schema fields such as `engine.version`, `engine.entry`, and `engine.model.params` are not consistently consumed.
6. Installation checks, static auth resolution, and real authentication validation do not have a clear lifecycle boundary.
7. Runner and `agent_judge` must share the same primitives without inferring an engine from a provider or duplicating compatibility branches.

## Model connection boundary

Harbor's recent model-connection work is a useful layering reference:

[...]

PR #211, opened 2026-08-19
refactor(agent): centralize resolved configuration

zpzjzj:
## Summary

- build one role-aware `ResolvedAgentConfig` after YAML, CLI, and credential loading
- resolve legacy slashed CLI model references once without mutating the loaded eval config, then pass the resolved value directly to adapters
- reuse the same resolver for `agent_judge` while inheriting the runner engine lifecycle, and report the resolved runner identity

## Compatibility

- preserves legacy `--model provider/name`, opaque slashed model IDs, YAML provider/name pairs, provider environment precedence, local-login delegation, and Qoder `auto` semantics
- lets an explicit CLI model bypass superseded YAML provider/name environment references while continuing to resolve active base URL and model params
- does not add `--provider`, protocol-aware nested credentials, capability validation, or a public judge-engine schema

## Testing

- `make verify`
- `make test`
- `make test-action`
- `make e2e` (all non-live paths passed; the local live Qoder CLI 0.2.0 run returned a successful response but did not expose token usage, so its existing non-zero token assertion failed)

Part of #196


internal/credential/agent_init.go
```diff
@@ -115,21 +137,25 @@ func parseJudgeModel(value string) (provider, model string) {
...
+		Engine:      in.engine.Name,
+		Version:     in.engine.Version,
+		Entry:       in.engine.Entry,
+		Provider:    in.provider,
+		Model:       in.model,
+		BaseURL:     in.baseURL,
+		Kwargs:      maps.Clone(in.engine.Kwargs),
+		ModelParams: maps.Clone(in.engine.Model.Params),
+		Custom:      custom,
 	}
```
JHWang-1997, line 160: **[P2] Deep-copy the custom-engine configuration**

`ResolvedAgentConfig` is documented as a stable value that cannot be changed by later mutations of `EvalConfig`, but assigning `custom` here preserves the original pointer and all of its nested maps, slices, and pointers. For a custom engine, changing `EvalConfig.Engine.Custom.Env`, `Local`, or `HTTP` after resolution therefore changes the already-resolved runner configuration; the judge also inherits the same pointer. Please deep-copy the complete custom configuration tree during resolution and add a regression test that mutates the source config afterward.

JHWang-1997 (changes requested):
Requesting changes for one immutability issue in the new resolved configuration. `Kwargs` and `ModelParams` are cloned, but the custom-engine configuration is still shared with the loaded eval config.

internal/credential/agent_init.go
zpzjzj: Fixed in 84d2eaa. The resolver now deep-copies the complete custom-engine configuration, including Env/Kwargs, Local.Args, HTTP headers/files/Required pointers, and nested YAML request-body maps/sequences. Judge resolution performs a separate copy from the runner config. Regression tests mutate the source EvalConfig and runner config after resolution and verify the resolved runner/judge values remain unchanged. `make verify` and `make test` pass.

JHWang-1997 (approved):
LGTM

Issue #206, opened 2026-08-18
feat: add --progress-file for machine-readable case progress events


JHWang-1997:
## Problem / motivation

`skill-up run` currently exposes case-level progress only through human-readable terminal output.

CI systems that need to periodically report evaluation progress must parse stdout, which is fragile because:

- the output format is intended for humans rather than machines;
- concurrent cases may interleave output;
- verbose logs may contain unrelated messages;
- wording or formatting changes can break downstream parsers.

`evaluator.ProgressObserver` already receives callbacks when cases start and complete, but these events are only forwarded to the terminal UI.

We need a stable, machine-readable progress stream that CI jobs can consume while an evaluation is still running.

## Proposed solution

Add an optional `--progress-file` flag to `skill-up run`:

```bash
skill-up run --progress-file ./progress.jsonl
```

When specified, skill-up should:

1. Create or truncate the target file when the run starts.
2. Append one JSON object per line as `ProgressObserver` callbacks occur.
3. Make each event immediately visible to readers so CI wrappers can tail or periodically read the file.
4. Preserve the existing terminal progress output.
5. Leave current behavior unchanged when the flag is not specified.

Example events:

```json
{"schema_version":"1","event":"case_started","time":"2026-08-18T10:00:00.002Z","iteration":1,"case_index":1,"case_total":2,"case_id":"case-1","configuration":"with_skill","title":"Basic flow"}
{"schema_version":"1","event":"case_completed","time":"2026-08-18T10:01:10.451Z","iteration":1,"case_index":1,"case_total":2,"case_id":"case-1","configuration":"with_skill","status":"PASS","pass_rate":1.0,"duration_ms":70449}
```

Suggested fields:

| Field | Description |
| --- | --- |
| `schema_version` | Event schema version, initially `"1"` |
| `event` | `case_started` or `case_completed` |
| `time` | Event timestamp in RFC 3339 format |
| `iteration` | Current iteration number |
| `case_index` | Existing observer case index |
| `case_total` | Total number of case tasks in the iteration |
| `case_id` | Case identifier |
| `configuration` | `with_skill` or `without_skill` |
| `title` | Case title, for `case_started` |
| `status` | `PASS`, `FAIL`, `ERROR`, or `SKIP`, for `case_completed` |
| `pass_rate` | Judge pass rate when available |
| `duration_ms` | Case duration, for `case_completed` |

[...]

PR #213, opened 2026-08-19
docs(proposal): define skill evaluation event log protocol

JHWang-1997:
## Summary

- define SUP-0005 as a general, versioned evaluation event log with JSONL as the v1 serialization
- define the normative event envelope, typed payload fields, lifecycle invariants, and schema-version consumer rules
- design an internal `evalevent` Go model with typed payloads, ordered Publisher fan-out, sticky per-sink errors, and a lifecycle state machine
- design one immutable invocation-wide TaskPlan so execution and CI progress share the same task identity and totals
- define the future `skill-up run --event-log <path>` CLI contract, path safety, dry-run behavior, startup/finalization order, and exit-error composition
- provide matching English and Chinese specifications and examples

## Motivation

CI integrations and other subscribers need structured evaluation state without parsing human-readable stdout. The design follows [zpzjzj's feedback on issue #206](https://github.com/alibaba/skill-up/issues/206#issuecomment-5336855301): keep stdout separate, treat progress as the first use case of a broader event log, and borrow Bazel BEP's core separation without copying its protobuf encoding or full DAG.

The additional design closes the implementation gaps identified in review: global progress must not reuse per-iteration `ProgressObserver` indices, event identity must be assigned before sink fan-out, payload wire types must be stable, and lifecycle completion must remain distinct from command success.

## Key design decisions

- `run_started`, iteration events, case events, and terminal `run_finished`
- invocation-wide `task_index`, `task_total`, and cumulative `completed_tasks`
- `COMPLETED` / `ERROR` / `CANCELLED` lifecycle status, independent of case-result exit behavior
- Publisher-owned `invocation_id`, contiguous `sequence_number`, timestamp, and terminal marker
- typed internal payload structs rather than `map[string]any`
- deterministic task planning shared by Runner, Evaluator, and lifecycle emission
- `--event-log` writes one invocation per file, truncates once, and remains separate from stdout
- event logging starts before credential loading and always attempts graceful terminal publication after startup
- no public `pkg/` API commitment in v1

## Impact

Documentation only. This PR does not add `--event-log`, change `ProgressObserver`, or modify runtime behavior. The implementation remains tracked in #206.

## Validation

- parsed every English and Chinese JSONL example with `jq`
- verified contiguous sequences, task totals, cumulative progress, terminal invariants, and one terminal event
- verified English and Chinese JSONL examples and Go design blocks match
- verified the illustrative Go event model parses with `gofmt`
- ran `git diff --check`
- ran `make verify`
- ran `make test` (`go test -race ./...`)


proposals/0005-skill-evaluation-event-protocol.md
```diff
@@ -0,0 +1,318 @@
...
+
+A task is one (`iteration`, `case_id`, `configuration`) tuple. Benchmark mode
+therefore creates two tasks from one source case. `task_id` is stable and
+unique within the invocation; a readable form is:
+
+```text
+iteration-1/case-1/with_skill
+```
+
+`task_index` and `task_total` describe planned tasks across the whole
```
zpzjzj: **🟠 Major: Define the global progress invariants**

The fields are sufficient for aggregate CI progress, but their lifecycle contract needs to be normative. Please specify that `run_started` is emitted exactly once before any case event; `task_total` is the final invocation-wide total after filtering, benchmark configuration expansion, and iteration expansion; `task_index` is stable, invocation-wide, and 1-based; and `completed_tasks` is invocation-wide and cumulative. For a normally finalized run, also require `completed_tasks == task_total == passed + failed + errored + skipped`. This matters because the current `ProgressObserver` indices reset for each iteration, so an implementation cannot forward them directly as protocol-level totals.

proposals/0005-skill-evaluation-event-protocol.md
```diff
@@ -0,0 +1,318 @@
...
+| `last_event` | boolean | Optional; present and `true` only on `run_finished` |
+| `payload` | object | Fields defined by the event type |
+
+The pair (`invocation_id`, `sequence_number`) is the event's stable identity
+for ordering and de-duplication. V1 intentionally has no event ID, parent ID,
+or graph relationship.
+
+### Initial Event Types
+
+| Event | Required payload fields |
```
zpzjzj, line 212: **🟠 Major: Specify normative payload wire types and constraints**

The registry currently lists field names but does not define their JSON types or essential domains. Since the migration section says changing a field's type or meaning requires a new major version, please add typed payload tables or a normative schema covering at least integer ranges, 1-based indexing, `pass_rate` as `0..1`, optional-vs-null behavior, `duration_ms`, and whether `report_dir` is an absolute path, relative path, or URI. Examples alone are not enough to prevent incompatible v1 producers and consumers.

proposals/0005-skill-evaluation-event-protocol.md
```diff
@@ -0,0 +1,318 @@
...
+
+JSONL is the v1 serialization. Files are the first proposed sink, not the
+protocol itself.
+
+### JSONL File Semantics
+
+The future JSONL file sink must:
+
+- create or truncate the requested path once at startup;
+- assign `sequence_number`, marshal the event, append `\n`, and write the
```
zpzjzj: **🟠 Major: Assign event identity before sink fan-out**

`sequence_number` should not be assigned by the JSONL sink. The proposal allows the same event to fan out to file, callback, OTLP, and other sinks; if one sink assigns the sequence, the other sinks cannot observe the same immutable event identity and ordering. Please make a publisher/dispatcher assign `invocation_id`, `sequence_number`, and `time` once under the ordering lock before fan-out, leaving each sink responsible only for transport or serialization.

proposals/0005-skill-evaluation-event-protocol.md
```diff
@@ -0,0 +1,318 @@
...
+  complete record under the same lock;
+- perform one `Write` per record and make completed records promptly visible;
+- avoid `fsync` on every event and avoid a delayed buffered flush;
+- guarantee only that each newline-terminated record is valid JSON.
+
+A concurrently reading process may observe a partial final write, especially
+after a crash or disk failure. Consumers therefore must not parse an
+unterminated trailing record as a complete event.
+
+### Failure Semantics
```
zpzjzj, line 783: **🟠 Major: Define terminal status and fan-out failure semantics**

Please define how `run_finished.status` maps to the command outcome. The proposal uses `SUCCESS` for normal finalization even though `skill-up run` may return a non-zero exit status when a `with_skill` task fails or errors; either use a lifecycle term such as `COMPLETED`, or expose an explicit command/result status or exit code. For fan-out, also specify that one failed sink does not prevent healthy sinks from receiving the current event, subsequent events, or the terminal event; sink failures should remain sticky per sink and be aggregated into the final command error.

zpzjzj (commented):
The current case lifecycle events are sufficient for the initial CI progress use case: a consumer can derive the total, started, completed, running, remaining, per-task results, aggregate results, and clean termination. Before freezing v1, the proposal should clarify the four protocol contracts below.

Review summary: 0 blockers, 4 major findings, 0 informational findings. Coverage: 3/3 changed files; the event contract, current runner/evaluator progress behavior, serialization examples, bilingual parity, and CI-only checks were reviewed. Both JSONL examples parse successfully, have monotonic sequence numbers, and contain exactly one terminal event. All current PR checks pass.

JHWang-1997:
Thanks for the detailed review. The latest revision now defines invocation-wide progress invariants, normative wire types and ranges, and rejection of unsupported protocol versions. Event identity is assigned once by the publisher before sink delivery.

Run status now uses COMPLETED, ERROR, and CANCELLED, and is explicitly separate from case outcomes and the command exit code. V1 is scoped to one synchronous JSONL sink, with sticky sink failures included in the final command error.

The English and Chinese examples have also been updated. Could you take another look when convenient?

proposals/0005-skill-evaluation-event-protocol.md
```diff
@@ -0,0 +1,915 @@
...
+- one global `task_index`, `task_id`, and `task_total` for every expanded task.
+
+Plan order is deterministic: iteration number ascending, filtered case order,
+then `with_skill` before `without_skill`. The canonical v1 task ID is:
+
+```text
+iteration-<iteration>/<case_id>/<configuration>
+```
+
+Case IDs are already validated not to contain path separators. Task start and
```
zpzjzj: **🟠 Major: Do not rely on a case-ID validation that does not exist**

The proposal says case IDs are already validated not to contain path separators, but `internal/config.Validator.ValidateCaseConfig` currently performs no such check and explicitly supplied IDs pass through unchanged. An implementation following this text could therefore produce the canonical slash-delimited `task_id` from an ID containing `/` or `\\`, despite the stated precondition.

**Suggestion**: either make separator rejection an explicit planning/config-validation requirement (including the compatibility impact on existing case IDs), or define an unambiguous escaping/encoding scheme and specify that consumers treat `task_id` as opaque.

proposals/0005-skill-evaluation-event-protocol.md
```diff
@@ -0,0 +1,915 @@
...
+6. Stop the heartbeat, enter `finalizing`, clear tasks that are no longer
+   active because of invocation cancellation/error without counting them as
+   completed, and choose `COMPLETED` when every planned task reached a case
+   terminal status,
+   `CANCELLED` for `context.Canceled` or `context.DeadlineExceeded`, and
+   `ERROR` for other invocation-level failures.
+7. Publish `run_finished` through `PublishLast`, close the file, and combine
+   evaluation, case-result, publisher, and close errors with `errors.Join`.
+
+Steps 6 and 7 use a fresh, bounded finalization context rather than an already
```
zpzjzj: **🟠 Major: Do not promise a timeout for synchronous file writes**

V1 specifies a synchronous JSONL sink, while the proposed bounded finalization context is expected to turn a finalization timeout into a sticky publisher error. A blocking `os.File.Write` does not observe context cancellation, so a stalled NFS/FUSE-backed file can keep steps 6–7 blocked beyond that bound and prevent the promised timeout error from being recorded.

**Suggestion**: either describe finalization as best-effort and remove the hard timeout guarantee for synchronous file writes, or define an actually interruptible/asynchronous write boundary and cover the stalled-write path in the future implementation tests.

zpzjzj (commented):
Follow-up review at `7a830db`: the previously raised protocol gaps are addressed, and those seven threads have been resolved.

Review summary: 0 blockers, 2 major findings, 0 informational findings. Coverage: 3/3 changed files; the event contract, current config validation and runner/evaluator behavior, serialization examples, bilingual parity, and CI-only checks were reviewed. The English and Chinese JSONL streams contain the same 16 events and satisfy the documented progress and terminal invariants. All 10 current checks pass.

Two remaining contract mismatches should be clarified before freezing v1.

proposals/0005-skill-evaluation-event-protocol.md
JHWang-1997: The current config validation does not reject path separators, so I removed that assumption. task_id is now an opaque ID that remains stable within an invocation. Consumers should correlate events using (invocation_id, task_id) and use the explicit iteration, case_id, and configuration fields instead of parsing the ID.

The task-N format is only an implementation example and is not part of the wire contract. The examples and test plan have also been updated. Artifact-path validation remains a separate command-level concern.

proposals/0005-skill-evaluation-event-protocol.md
JHWang-1997: A context deadline cannot reliably interrupt a synchronous os.File.Write, so the bounded-finalization guarantee has been removed.

Finalization still uses a fresh context to attempt to write the final snapshot and run_finished after cancellation, but this is now explicitly best-effort. CI remains responsible for the outer job or process timeout. Interruptible or asynchronous sinks can be considered separately if needed.

The sink contract, command flow, failure semantics, and drawbacks have been updated accordingly.

proposals/0005-skill-evaluation-event-protocol.md
```diff
@@ -0,0 +1,937 @@
...
+  "protocol_version": 1,
+  "event_version": 1,
+  "sequence_number": 3,
+  "invocation_id": "018f8f20-7a7d-7d90-a192-4f5ec8f07a2a",
+  "time": "2026-08-19T10:01:10.451Z",
+  "event": "case_completed",
+  "attributes": {
+    "com.alibaba.aone.eval_task_id": "123456"
+  },
+  "payload": {}
```
zpzjzj: **🟠 Major: Make the envelope example valid for its declared event**

This concrete envelope declares `event: "case_completed"` but uses an empty `payload`. The registry later requires `task_id`, `iteration`, `case_id`, `configuration`, `task_index`, `task_total`, `title`, `completed_tasks`, `status`, and `duration_ms`, and the test plan says typed payload constraints will be validated. As written, a consumer or schema test following the proposal must reject this example.

Please either populate a valid `case_completed` payload or make the envelope explicitly schematic without naming a concrete event, and mirror the correction in the Chinese proposal.

zpzjzj (commented):
Re-review at `ebcdf248`: **0 Blockers, 1 Major, 1 Info**.

Coverage: all 3/3 changed files; the event envelope, registry constraints, lifecycle/failure semantics, JSONL examples, test plan, and English/Chinese mirrors were checked. The 16-event bilingual streams remain aligned, and all current CI checks pass. Implementation-only and credential-gated paths are not applicable to this proposal-only change.

**🔵 Info:** In the heartbeat section, “the latest snapshot unchanged” is ambiguous because `elapsed_ms` is expected to advance and the examples update it from 10007 to 40007. Please clarify in both language versions that the phase and counters remain unchanged while `elapsed_ms` is refreshed.

Overall: ⚠️ conditionally mergeable after the Major below is addressed.

JHWang-1997:
The heartbeat wording has been clarified in both language versions. A periodic heartbeat keeps the current phase and counters unchanged while refreshing elapsed_ms. The lifecycle-emitter description and stream introduction now use the same wording.

proposals/0005-skill-evaluation-event-protocol.md
JHWang-1997: The envelope example now includes a complete case_completed payload with all required fields and valid values. I also changed sequence_number to 7 so the concrete example is consistent with the required preceding lifecycle events. The same change is mirrored in the Chinese proposal.

proposals/0005-skill-evaluation-event-protocol.md
```diff
@@ -0,0 +1,951 @@
...
+
+### Sink and Transport Model
+
+Event production targets an extensible sink boundary instead of adding more
+positional parameters to `ProgressObserver`:
+
+```go
+type EventSink interface {
+    Publish(context.Context, Event) error
+}
```
zpzjzj, line 640: **🟠 Major: Define ownership of sink closure**

The only sink contract shown here exposes `Publish`, but the publisher later promises `Close`, sticky close errors through `Err`, and final aggregation of the underlying file-close failure. An implementation receiving only this `EventSink` cannot satisfy those guarantees without an additional undocumented interface or concrete-type dependency. Please add an explicit close operation to the sink boundary (or compose it with a closer) and specify who invokes it exactly once and how its error is retained and aggregated.

proposals/0005-skill-evaluation-event-protocol.md
```diff
@@ -0,0 +1,951 @@
...
+
+`runEval` should be refactored around one explicit finalization path:
+
+1. Parse flags and attributes, load and validate configuration, and apply
+   filters and overrides.
+2. Reject an incompatible dry-run/event-log combination, then build the
+   immutable task plan.
+3. Validate and open the requested JSONL sink, create the publisher and
+   lifecycle emitter, and publish `run_started` plus the initial `preparing`
+   progress snapshot. A first-record failure uses the startup behavior above.
```
zpzjzj: **🟠 Major: Define failure of the initial progress write**

`run_started` completes event-log startup, but this immediately following `preparing` snapshot is still written before credentials are loaded and before the best-effort-finalization region begins. If `run_started` succeeds and this second record fails, the sticky failure disables the sink so no terminal marker can be written, while the command contract does not say whether agent execution must abort or may continue. Please define this path explicitly and include it in the startup/failure tests.

proposals/0005-skill-evaluation-event-protocol.md
```diff
@@ -0,0 +1,951 @@
...
+The v1 command contract is:
+
+- both flags are optional and have no config-file or
+  environment-variable equivalent;
+- an empty value and `-` are rejected; stdout remains human-readable and is
+  never an event transport;
+- a relative path is resolved against the process working directory;
+- the parent directory must already exist; the command creates the file with
+  mode `0600` subject to the process umask, or truncates an existing file once
+  without changing that file's existing permissions;
```
zpzjzj: **🟠 Major: Make the permission guarantee portable**

This is an unconditional POSIX `0600`/umask contract, while skill-up also supports native Windows. Go file creation on Windows cannot enforce POSIX mode or umask semantics, so the proposed implementation and permission test cannot satisfy this wording on every supported host. Please either define an equivalent Windows ACL behavior or explicitly scope the mode guarantee and its tests to POSIX, while documenting the Windows behavior.

proposals/0005-skill-evaluation-event-protocol.md
```diff
@@ -0,0 +1,951 @@
...
+  discovery, or task-plan updates in v1.
+- Streaming prompts, model responses, transcripts, credentials, or artifact
+  contents.
+- Reproducing the full Bazel BEP graph or protobuf schema.
+- Adopting CloudEvents as the v1 wire representation. A future adapter can map
+  records without changing this ordered-log contract.
+
+## Requirements
+
+- Each complete record is one valid JSON object terminated by `\n`.
```
zpzjzj: **🟠 Major: Require unique JSON member names**

Requiring a `valid JSON object` does not make duplicate member names unambiguous: parsers may reject them, retain the first value, retain the last value, or expose all values. Duplicate envelope fields such as `sequence_number`, `event`, or `last_event` can therefore make different conforming consumers disagree about identity, ordering, and stream completion. Please require member names to be unique at every object level and classify a complete record containing duplicates as malformed/corrupt; the consumer/schema tests should cover this case.

zpzjzj (commented):
Re-review at `8592163`: **0 Blockers, 4 Major findings, 0 informational findings published**.

Coverage: all 3/3 changed files; the event envelope, lifecycle and failure semantics, sink ownership, JSONL parsing rules, cross-platform CLI behavior, current runner/evaluator integration points, bilingual examples, and test plan were checked. The English and Chinese examples remain aligned, all 16 JSONL records parse and satisfy the documented sequence/terminal invariants, `make test` passes locally, and all 10 current PR checks pass.

The previous envelope-payload and heartbeat wording issues are fixed. The four remaining contract gaps below should be resolved before freezing v1. Overall: conditionally mergeable after these Major findings are addressed.

proposals/0005-skill-evaluation-event-protocol.md
JHWang-1997: Updated in 33fccf3. EventSink now includes Close, and a successfully constructed Publisher owns the Sink. Callers do not close it directly. Publisher.Close is idempotent, closes the underlying Sink at most once, and is required on every controlled path, including after a publication failure. It retains and aggregates sticky publication errors with close-time errors. The test plan now covers idempotent close, close after publication failure, and combined publication/close errors.

proposals/0005-skill-evaluation-event-protocol.md
JHWang-1997: Updated in 33fccf3. Event-log startup now completes only after both run_started and the initial preparing run_progress snapshot are published successfully. Failure of either publication closes the Publisher and exits before credentials, Agent construction, or evaluation; it does not attempt run_finished. A second-record failure may leave only a complete run_started prefix or a partial trailing write, so the file is explicitly incomplete. The test plan covers the two startup publications separately, one underlying Sink close, and no final-event attempt.

proposals/0005-skill-evaluation-event-protocol.md
JHWang-1997: Updated in 33fccf3. The 0600 and umask guarantee is now scoped to POSIX. On Windows, a new file uses the default security descriptor inherited from the parent directory, with no 0600-equivalent ACL guarantee, and v1 does not modify ACLs. Truncating an existing file preserves its mode or ACL. The test plan scopes the 0600 assertion to POSIX and keeps create/truncate behavior covered across platforms.

proposals/0005-skill-evaluation-event-protocol.md
JHWang-1997: Updated in 33fccf3. Member names must now be unique in every JSON object after string escapes are decoded. A complete newline-terminated record containing duplicate names at any depth is malformed and corrupts the stream, and consumers must reject it before interpreting the envelope or payload. The test plan covers envelope fields, nested objects, and escape-equivalent member names.

zpzjzj (approved):
LGTM

PR #217, opened 2026-08-20
refactor(agent): declare adapter capabilities

zpzjzj:
## Summary

- declare adapter protocol, model, endpoint, and kwarg capabilities
- resolve requested versus applied configuration once for runner and judge, surfacing unsupported explicit settings before execution
- record credential-free requested/applied configuration in sessions, telemetry, JSON, HTML, and Markdown while preserving legacy `model_name` semantics
- record observed model identity only when the agent explicitly reports it; local/default selection remains unknown
- document the future boundary between provider configuration, a protocol-aware resolved model connection, adapter materialization, and agent-native configuration

## Configuration semantics

- `requested_configuration`: values selected through YAML, CLI, environment, and credential precedence
- `applied_configuration`: values skill-up forwards to the CLI after adapter normalization; local CLI settings may still override them
- `observed_configuration`: emitted only when every runner session explicitly reports the same model; per-case observations remain in `observed_model`

When Codex cannot apply a requested custom provider, its applied provider, model, endpoint, and provider-scoped credential are all omitted before falling back to local Codex settings.

Custom engines consume the top-level provider/model/API-key identity, but custom endpoints and parameters remain under `engine.custom.http.url` and `engine.custom.kwargs`; unused top-level `base_url` or `engine.kwargs` are now surfaced rather than reported as applied.

Capability warnings do not include configured kwarg values, preventing an accidentally supplied credential from being persisted in logs or `result.json`.

No auth probe or extra model request is introduced, so existing local CLI login remains valid without consuming tokens merely to inspect state.

## Model connection follow-up

The design follows Harbor's useful layering (`ProviderAccess` -> `ResolvedModelConnection` -> native agent config) without copying its model-based provider inference or credential alias fan-out. Protocol-aware connection resolution, adapter materialization, `--provider` compatibility, and native configuration remain staged follow-ups in #196 rather than expanding this PR.

## Compatibility

- no eval schema or CLI flag changes
- legacy provider/model parsing and top-level `model_name` semantics are preserved
- installation/version lifecycle and protocol-aware connection resolution remain follow-ups

## Testing

- `make verify`
- `make test`
- `make e2e`
- `go test -tags e2e -v ./e2e -run TestPipeline_ReportsRequestedAppliedAndUnknownObservedModel -count=1`
- two fixed-SHA self-review passes; the first found and fixed custom capability overclaiming and raw invalid-kwarg warnings, and the second found no actionable defects

One separate self-review-triggered E2E rerun observed the existing local Qoder behavior where a successful run reports zero token usage; the full pre-push `make e2e` run passed.

Part of #196


internal/agent/capabilities.go
```diff
@@ -0,0 +1,254 @@
...
+		if slices.Contains(supportedQoderModels, requested) {
+			return requested
+		}
+		params.Warnings = appendUniqueWarning(params.Warnings, fmt.Sprintf(
+			"engine %q does not support model %q; the model override is omitted and local Qoder settings will be used",
+			params.Engine, requested,
+		))
+		return ""
+	case ModelPolicyCodexProvider:
+		if reason := codexCustomProviderUnavailableReason(params.Provider, params.BaseURL); reason != "" {
```
JHWang-1997, line 167: **Track the effective provider independently from the requested provider**

When a non-OpenAI Codex provider is unusable, this code clears only `EffectiveModel`; `Provider` remains unchanged and is later copied into `effective_configuration.provider`. However, `runProviderConfig` drops that provider and Codex runs with its local/default provider configuration, so the report can disagree with the actual execution.

For example:

```yaml
engine:
  name: codex
  model:
    provider: dashscope
    name: qwen3.6-plus
    # base_url is missing
```

The resulting report can contain:

```json
{
  "requested_configuration": {
    "provider": "dashscope",
    "model": "qwen3.6-plus"
  },
  "effective_configuration": {
    "protocol": "openai",
    "provider": "dashscope"
  }
}
```

But the generated Codex command contains neither `model_provider=dashscope` nor `-m qwen3.6-plus`; it delegates both choices to local/default Codex settings. If the requested model is empty, the early return above also skips the provider warning entirely. In addition, the resolved provider API key may still be injected as `OPENAI_API_KEY` after the provider configuration has been dropped, potentially sending the wrong credential to the default endpoint.

Please resolve and record `EffectiveProvider` independently (and ideally the effective endpoint/credential routing as well). When falling back to local Codex settings, `effective_configuration.provider` and `.model` should both be empty/default, and credentials belonging to the dropped provider should not be injected.

JHWang-1997 (commented):
The requested/effective split is useful, but the effective provider identity can diverge from the provider actually used by Codex when provider configuration falls back to local defaults. The inline comment includes a concrete example and suggested resolution.

internal/agent/capabilities.go
zpzjzj: Fixed in e8ac6ff. Requested and applied provider identities are now resolved independently. When Codex cannot apply a custom provider, the applied provider, model, endpoint, and provider-scoped credential are all omitted, so the rejected credential is not injected into OPENAI_API_KEY while Codex falls back to local settings. The warning is also emitted when the requested model is empty. Added resolver and factory/env regression coverage.

JHWang-1997 (approved):
LGTM

PR #220, opened 2026-08-26
fix(qwen-code): exclude hidden thoughts from transcripts

chiga0:
## Summary

Qwen Code session JSONL can contain assistant text parts annotated with `thought: true`. The parser currently treats these hidden reasoning parts as visible assistant messages, so they can replace or contaminate `FinalMessage` and downstream transcript consumers.

This change:
- models the `thought` marker and skips those text parts
- preserves visible assistant text, tool calls, and tool results
- adds a regression test covering the full sequence

## Validation

- `make verify`
- `go test ./internal/agent -run Qwen -count=1`
- `env -u QODER_PERSONAL_ACCESS_TOKEN -u OPENAI_API_KEY go test -race ./internal/agent ./internal/runtime`

The initial full `make test` run was otherwise green but hit one host-environment credential test and one timing-sensitive runtime test; both affected packages passed cleanly when rerun with credential variables removed.

### DIFF
--- a/internal/agent/qwen_code.go
+++ b/internal/agent/qwen_code.go
@@ -318,6 +318,7 @@ type qwenSessionMessage struct {
 
 type qwenSessionPart struct {
 	Text             string                `json:"text,omitempty"`
+	Thought          bool                  `json:"thought,omitempty"`
 	FunctionCall     *qwenFunctionCall     `json:"functionCall,omitempty"`
 	FunctionResponse *qwenFunctionResponse `json:"functionResponse,omitempty"`
 }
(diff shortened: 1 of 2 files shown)


JHWang-1997 (approved):
Reviewed the Qwen session parsing change. Qwen stores hidden reasoning as text parts marked thought: true, which the current parser exposes as normal assistant messages. Skipping them limits the transcript and final message to observable output while preserving visible responses, tool activity, and token accounting.

Our agent judge can read the transcript, so historical results that used leaked reasoning may change. This is an intended correction and aligns Qwen with Claude, Qoder, and Codex.

Targeted race tests pass, with no blockers found. Please add an Unreleased changelog entry and run the normal CI checks, since GitHub currently shows only the CLA check.

LGTM.

PR #226, opened 2026-09-02
refactor(agent): materialize protocol-aware model connections

zpzjzj:
## Summary

- resolve one model connection after the adapter protocol is known
- make adapter construction consume that applied connection instead of reinterpreting flattened credential fields
- preserve agent-local authentication and routing, including explicit-empty overrides
- reuse the runner connection for inherited agent judges while resolving different providers independently

## Why this scope

This is PR 4b of #196 and builds on #221 and #222. It migrates adapter materialization without adding native agent config, installation or version checks, or an independent judge-engine schema.

## Compatibility

- existing v1alpha1 YAML, CLI, and GitHub Action forms remain unchanged
- flat provider credentials still work
- nested OpenAI and Anthropic endpoints now select by adapter protocol
- missing credentials and endpoints continue to delegate to local agent state
- rejected Codex provider config does not leak into local fallback

## Validation

- `make verify`
- `make test`
- `make test-action`
- `make build`
- `make e2e`
- `go test -race ./internal/credential ./internal/agent ./internal/evaluator ./internal/runner ./internal/cli`

Part of #196.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Agent adapters now consume one protocol-aware model connection selected from
+  provider configuration. OpenAI- and Anthropic-compatible endpoints under the
+  same provider no longer pass through the legacy flattened credential view,
+  while absent credentials and endpoints continue to delegate to agent-local
+  login and routing.
+
 ## [0.10.0] - 2026-09-01
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   while absent credentials and endpoints continue to delegate to agent-local
   login and routing.
 
+### Fixed
+- Agent judges selecting a different provider no longer inherit runner credentials
+  or endpoints, preventing judge keys from being sent to a runner endpoint.
+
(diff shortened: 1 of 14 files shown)


JHWang-1997 (changes requested):
When judge.model selects a different provider, the runner-sourced connection values are still carried into the judge connection.

For example, with a runner endpoint at runner.example.test and a judge provider configured with judge-key and judge.example.test, the final Claude environment contains judge-key but still uses runner.example.test as ANTHROPIC_BASE_URL. This can send the judge credential to the wrong endpoint.

I confirmed that this behavior also exists on the base commit, so it is not introduced by this PR. However, this PR makes AppliedConnection authoritative and states that different judge providers are resolved independently. Could we resolve the judge connection entirely from its own provider and protocol, and add a regression test covering different runner and judge providers?

zpzjzj:
Thanks for catching this. Fixed in f51eb56: runner key and endpoint fallback now requires the same provider namespace. A judge selecting a different provider resolves its own connection using the inherited adapter protocol; missing values no longer fall back to the runner connection. Same-provider inheritance and agent-local delegation remain unchanged.

Added regression coverage for flat and protocol-specific credentials, missing keys/endpoints, explicit-empty values, and nil-resolver inheritance. The cross-provider tests cover both Anthropic and OpenAI-compatible adapters through factory construction and the generated credential environment.

Validation passed: make verify, make test (race-enabled), make test-action, make build, and make e2e (quick mode; credential-dependent cases remain gated).

JHWang-1997 (approved):
LGTM

PR #228, opened 2026-09-02
refactor(evaluator): introduce execution plans

JHWang-1997:
## Summary

- build one deterministic invocation-wide plan across iterations, cases, and benchmark configurations
- execute the planned tasks through preserved Runner.Evaluate and Evaluator.EvaluateAll compatibility entry points
- add optional task and iteration lifecycle observer boundaries for the later event-log integration

## Compatibility

User-visible changes: none.

This PR does not add CLI flags or event files. It preserves existing cancellation scheduling, terminal progress, report handling, result ordering, and exit behavior.

## Validation

- make fmt
- make verify
- make test
- make e2e

Part of #206.

PR #231, opened 2026-09-02
feat(cli): add evaluation event logs

JHWang-1997:
## Summary

- add opt-in --event-log output and repeatable --event-attribute metadata
- connect the deterministic execution plan and runner observers to the evaluation event lifecycle
- validate output paths before opening the JSONL file and preserve graceful finalization on cancellation
- document the CLI contract and add unit and end-to-end coverage

## Compatibility

The existing evaluation path is unchanged when --event-log is omitted. No event objects, files, or heartbeat work are created in that mode, and existing stdout, reports, and exit behavior are preserved.

When event output is enabled, sink errors are returned with the command result according to SUP-0005.

## Testing

- make fmt
- make verify
- env -u QODER_PERSONAL_ACCESS_TOKEN make test
- env -u QODER_PERSONAL_ACCESS_TOKEN make GOFLAGS="-buildvcs=false -parallel=1" e2e

Builds on #228 and #229.
Addresses #206.

### DIFF
--- a/e2e/event_log_signal_unix_test.go
+++ b/e2e/event_log_signal_unix_test.go
@@ -67,10 +67,24 @@ input:
 `)
 
 	eventPath := filepath.Join(dir, "events.jsonl")
+	outputDir := filepath.Join(dir, "artifacts")
+	futureArtifacts := []string{
+		filepath.Join(outputDir, "iteration-2", "keep.txt"),
+		filepath.Join(outputDir, "iteration-3", "keep.txt"),
+	}
+	for _, artifactPath := range futureArtifacts {
+		if err := os.MkdirAll(filepath.Dir(artifactPath), 0o755); err != nil {
+			t.Fatalf("create future iteration directory: %v", err)
+		}
+		if err := os.WriteFile(artifactPath, []byte("keep"), 0o600); err != nil {
+			t.Fatalf("write future iteration artifact: %v", err)
+		}
+	}
 	cmd := exec.Command(
 		binaryPath,
 		"run", filepath.Join(dir, "evals", "eval.yaml"),
-		"--output-dir", filepath.Join(dir, "artifacts"),
+		"--iteration", "3",
+		"--output-dir", outputDir,
 		"--event-log", eventPath,
 	)
 	cmd.Dir = dir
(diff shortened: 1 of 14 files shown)


zpzjzj (commented):
Reviewed all 8 changed files, including CLI/runner/lifecycle integration, path validation, serialization, tests, and documentation. This review contains 0 blockers, 1 major finding, and 0 informational findings. `make test` (race-enabled), `make verify`, and the default E2E suite passed. Credential-gated full LLM/OpenSandbox paths were not validated.

internal/cli/run_events.go
```diff
@@ -0,0 +1,507 @@
...
+		evalCfg,
+		loader,
+		resolver,
+		runnerConfig,
+		ag,
+		executionPlan,
+		evaluateOpts,
+		stream.adapter,
+	)
+	finalErr := stream.finish(context.WithoutCancel(cmd.Context()), eventRunStatus(cmd.Context(), evalErr))
```
zpzjzj: **[P2] Connect process signals to invocation cancellation**

This finalization path is not reached when users cancel the actual CLI with SIGINT or SIGTERM: `ExecuteArgs` in `internal/cli/root.go` creates the command context from `context.Background()`, and no signal handler cancels it. I reproduced this with the mock engine by sending each signal after `case_started`: the process exited by signal and the log ended at the sixth record (`run_progress`), with neither `run_finished`/`CANCELLED` nor `last_event: true`. Thus the new graceful-cancellation handling currently works only for manually cancelled contexts, not the CLI entry point used by users and CI. Please connect supported termination signals to the event-enabled invocation context, let evaluation unwind into this finalizer, and add a subprocess-level cancellation test. Keep final publication on the uncancelled context and preserve the existing behavior when event logging is disabled.

internal/cli/run_events.go
```diff
@@ -0,0 +1,507 @@
...
+	}
+	if canonicalPath == workspacePath {
+		return "", fmt.Errorf("--event-log path %q conflicts with the evaluation workspace", path)
+	}
+	for _, iteration := range plan.TaskPlan.Iterations {
+		iterationPath, err := canonicalizePotentialPath(filepath.Join(plan.WorkspaceDir, fmt.Sprintf("iteration-%d", iteration.Number)))
+		if err != nil {
+			return "", fmt.Errorf("resolve iteration %d workspace: %w", iteration.Number, err)
+		}
+		if pathWithin(iterationPath, canonicalPath) {
```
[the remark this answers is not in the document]
JHWang-1997: Fixed in f33051b. Path containment now compares existing directory identities with os.SameFile before falling back to lexical comparison, so case aliases on a case-insensitive filesystem are rejected before the event sink is opened. I also added a regression test that exercises Iteration-1 versus iteration-1 on case-insensitive filesystems.

internal/cli/run_events.go
JHWang-1997: Fixed in f33051b. Event-enabled runs now install a scoped SIGINT/SIGTERM cancellation context, allowing evaluation to unwind through the existing uncancelled finalizer and publish run_finished with status CANCELLED and last_event=true. The signal context is installed only on the event-log path and restored afterward, so runs without --event-log keep their existing signal behavior. A subprocess E2E test now covers both SIGINT and SIGTERM.

internal/cli/run_events.go
```diff
@@ -0,0 +1,550 @@
...
+	ui.Blank()
+	ui.Stepf("🚀", "Running evaluation (%d cases)", plan.CaseCount)
+
+	evaluateOpts.Observer = &uiProgressObserver{}
+	evaluateOpts.TaskObserver = adapter
+	evaluateOpts.IterationObserver = adapter
+
+	run := runner.NewRunner(evalCfg, loader, resolver, runnerConfig)
+	logging.DebugContextf(cmd.Context(), "Runner: starting evaluation for %d case(s) with agent %s", plan.CaseCount, evalCfg.Engine.Name)
+	results, err := run.EvaluatePlan(cmd.Context(), plan, ag, evaluateOpts)
```
zpzjzj, line 198: **[P2] Stop scheduling iterations and cleaning workspaces after cancellation**

The new signal context reaches `Runner.EvaluatePlan`, but its iteration loop never checks cancellation before `setupWorkspace`, and `defaultEvaluator.EvaluatePlan` likewise continues scheduling tasks. I reproduced this with `--event-log ... --iteration 3` and a slow mock case: after sending SIGINT at the first `case_started`, the log still contained `iteration_started` and `case_started` for iterations 2 and 3. Both later iterations deleted their pre-existing artifacts through `setupWorkspace`/`os.RemoveAll`, then produced ERROR results before the final CANCELLED event. Capturing the signal must stop new work, not merely change the eventual run status. Please check cancellation before scheduling further tasks/iterations and before workspace cleanup, let already-started work unwind, and add a multi-iteration cancellation regression that verifies untouched future artifacts are preserved.

internal/cli/run_events.go
```diff
@@ -0,0 +1,550 @@
...
+	)
+	finalErr := stream.finish(context.WithoutCancel(cmd.Context()), eventRunStatus(cmd.Context(), evalErr))
+	if evalErr != nil {
+		return errors.Join(evalErr, finalErr)
+	}
+
+	ui.Separator()
+	passed, failed, errored := countResultStatus(results)
+	ui.Summary(passed, failed, errored)
+	return errors.Join(exitStatusError(cmd.ErrOrStderr(), results), finalErr)
```
zpzjzj, line 165: **[P2] Return invocation cancellation independently of baseline grading**

`eventRunStatus` observes `cmd.Context().Err()` and writes CANCELLED, but that cancellation is not included in the returned error. `Runner.EvaluatePlan` can return results with a nil error after cancellation, and `exitStatusError` deliberately ignores `without_skill` errors. With `--baseline --parallelism 1`, I let the with-skill case PASS and sent SIGINT when the without-skill case started: the CLI wrote `run_finished` with `status: CANCELLED` (one PASS and one ERROR) yet exited with code 0. CI therefore treats a cancelled invocation as successful, contrary to the protocol's non-zero requirement for CANCELLED. Please propagate the invocation context error into the command result independently of case-result filtering, while preserving uncancelled final publication, and cover cancellation during the baseline phase in a subprocess test.

zpzjzj (commented):
Re-reviewed all 9 changed files at f33051b, including CLI flags/configuration, runner/evaluator scheduling and cleanup, lifecycle serialization, reports/exit status, tests, and documentation. The two previously reported issues are fixed. This review contains 0 blockers, 2 major findings (P2), and 0 informational findings.

`make test` (race-enabled) and `make verify` passed. Normal event-lifecycle and SIGINT/SIGTERM subprocess E2E tests passed on rerun; an initial signal test timed out before startup, and a separate runtime timing test passed on rerun. Both findings below were reproduced with the actual CLI and mock engine. Credential-gated live-engine/OpenSandbox paths were not exercised.

internal/cli/run_events.go
JHWang-1997: Fixed in e9c0150. Task scheduling now observes cancellation while waiting for a concurrency slot and rechecks it before launching work, so unscheduled tasks emit no events. The runner checks cancellation before each iteration and again before removing the iteration workspace, preserving future iteration artifacts. Added evaluator and runner unit tests plus a multi-iteration SIGINT/SIGTERM subprocess regression that verifies iterations 2 and 3 are not started and their pre-existing artifacts remain untouched.

internal/cli/run_events.go
JHWang-1997: Fixed in e9c0150. The event-enabled run now combines the evaluation error with the invocation context error before finalization, uses that combined error for lifecycle status, and returns it independently of baseline result filtering. Final publication still uses an uncancelled context. Added a subprocess regression where the with-skill task passes, SIGINT is sent after the without-skill task starts, and the CLI records CANCELLED and exits with code 1.

e2e/event_log_signal_unix_test.go
```diff
@@ -0,0 +1,376 @@
...
+	if err := cmd.Start(); err != nil {
+		t.Fatal(err)
+	}
+	waitCh := make(chan error, 1)
+	go func() {
+		waitCh <- cmd.Wait()
+		close(waitCh)
+	}()
+
+	if err := waitForEventOrProcessExit(eventPath, "case_started", waitCh, 10*time.Second); err != nil {
```
zpzjzj: **[P2] Avoid a load-sensitive startup deadline in the signal E2E tests**

The new subprocess tests allow only 10 seconds for the CLI to emit `case_started` (and the baseline variant uses the same bound around line 229). This deadline has now failed under concurrent build/test load in separate review runs: one run timed out on SIGINT, and another timed out on SIGINT, SIGTERM, and the baseline wait, all before any stdout/stderr appeared; isolated reruns passed consistently. That makes the repository's required E2E gate dependent on host load rather than behavior. Please use a more generous process-start deadline or a deterministic readiness mechanism, and apply the same fix to both waits.

zpzjzj (commented):
Re-reviewed all 14 changed files at e9c0150, including cancellation scheduling, workspace cleanup, result and exit semantics, lifecycle serialization, tests, and documentation. The two previously reported cancellation issues are fixed. This review contains 0 blockers, 1 major finding (P2), and 0 informational findings.

`make test` (race-enabled), `make verify`, Windows cross-compilation, and the targeted event-log E2E scenarios passed, including SIGINT, SIGTERM, future-artifact preservation, baseline cancellation with a non-zero exit, and the normal lifecycle. The full credential-gated E2E path was not validated because the live Qoder client was not authenticated.

e2e/event_log_signal_unix_test.go
JHWang-1997: Fixed in 97e5ce3. Both waits for case_started now use a shared 60-second process startup timeout. The post-signal exit checks remain at 10 seconds so they still verify that cancellation completes promptly. The targeted signal E2E tests and go test ./... pass after syncing with current main.

zpzjzj (approved):
LGTM

PR #233, opened 2026-09-04
feat(agent): add static runtime preflight

zpzjzj:
## Summary

- run binary availability and static version checks before each case without probing authentication
- enforce exact configured CLI versions for supported isolated adapters while never modifying `runtime: none` hosts
- record requested, applied, and observed versions in JSON, Markdown, and HTML reports

## Compatibility

- local login remains delegated to the CLI; preflight makes no authentication or model request
- QoderCLI versions are observed, while `engine.version` remains unsupported and is ignored
- historical YAML, CLI override, and GitHub Action behavior remains compatible

## Validation

- `make verify`
- `make test`
- `make test-action`
- `make build`
- `go test -tags e2e -v -parallel 1 ./e2e`

Part of #196 (PR 5).

### DIFF
--- a/internal/config/validator.go
+++ b/internal/config/validator.go
@@ -81,6 +81,9 @@ func (v *Validator) ValidateEvalConfig(cfg *EvalConfig) error {
 	if cfg.Engine.Name == "" {
 		errs = append(errs, "engine.name is required")
 	}
+	if cfg.Engine.Version != "" && agentkind.SupportsVersion(cfg.Engine.Name) && !agentkind.IsExactVersion(cfg.Engine.Version) {
+		errs = append(errs, fmt.Sprintf("engine.version for %q must be an exact semantic version, got %q", cfg.Engine.Name, cfg.Engine.Version))
+	}
 
 	// engine.custom validation is deferred to ResolveCustomEngineConfig, which
 	// runs after CLI overrides settle the final engine name.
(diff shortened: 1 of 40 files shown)


internal/config/validator.go
```diff
@@ -81,6 +81,9 @@ func (v *Validator) ValidateEvalConfig(cfg *EvalConfig) error {
 	if cfg.Engine.Name == "" {
 		errs = append(errs, "engine.name is required")
 	}
+	if cfg.Engine.Version != "" && agentkind.SupportsVersion(cfg.Engine.Name) && !agentkind.IsExactVersion(cfg.Engine.Version) {
```
JHWang-1997: Before this change, a v1alpha1 config using engine.version values such as latest or ^1.2.0 passed validation and ran with a warning. The same config now fails for Claude Code, Codex, and Qwen Code, which seems inconsistent with the compatibility statement and the acceptance criteria in #196. Could we keep the previous warn-and-ignore behavior for non-exact versions, or explicitly document this as a breaking change with a migration path?

zpzjzj: Good catch. I restored backward compatibility: non-exact selectors such as `latest` and `^1.2.0` are now retained as requested configuration, warned and ignored, and omitted from applied configuration. Only exact semantic versions are passed to supported installers and enforced by preflight; observed versions still come from the static `--version` check. I also added validation, CLI override, capability, and report regression coverage and updated the docs.

JHWang-1997 (approved):
The compatibility issue is resolved. Non-exact version selectors remain accepted, are reported as requested configuration, and are omitted from the applied configuration with a warning. Exact versions still drive installation and preflight validation. The regression coverage looks sufficient. No further concerns from me.

PR #242, opened 2026-09-11
fix(agent): isolate qodercli from inherited SDK mode

### LINKED ISSUE #240: [Bug] qodercli engine inherits Qoder Agent SDK environment and fails with sdk_invalid_args
## Description

When `skill-up` is launched from a Qoder Agent SDK-hosted environment, such as a shell tool running inside Qoder Work, evaluations using the `qodercli` engine fail before the agent starts.

The parent environment contains:

```text
QODER_AGENT_SDK_ENTRYPOINT=sdk-ts
```

`skill-up` inherits this variable and forwards it to the qodercli child process. The child qodercli therefore enters Agent SDK mode, while the skill-up adapter invokes it using the ordinary headless CLI protocol.

This produces the following error:

```text
sdk_invalid_args: Agent SDK entrypoint env is set but required flags are missing.
Expected --print --input-format stream-json --output-format stream-json,
got print=true, inputFormat=undefined, outputFormat=json
```

Installing Qoder Work alone does not trigger the problem. It occurs only when the `skill-up` process actually inherits the SDK environment variable.

## Steps to reproduce

1. Use any valid evaluation configured with the qodercli engine:

   ```yaml
   engine:
     name: qodercli
   ```

2. Set the environment variable normally present in a Qoder Agent SDK-hosted process:

   ```bash
   export QODER_AGENT_SDK_ENTRYPOINT=sdk-ts
   ```

3. Run the evaluation:

   ```bash
   skill-up run ./evals/eval.yaml
   ```

4. Observe that qodercli exits during argument validation with `sdk_invalid_args`.

The qodercli behavior can also be reproduced independently using the same command shape generated by the adapter:

```bash
QODER_AGENT_SDK_ENTRYPOINT=sdk-ts \
qodercli \
  --permission-mode=bypass_permissions \
  --output-format json \
  -p 'reply only OK'
```

## Expected behavior

The qodercli adapter should start a fresh ordinary headless qodercli process regardless of whether `skill-up` was itself launched from an SDK-hosted parent process.

Host-only SDK control variables should not change the protocol selected by a nested qodercli invocation.

## Actual behavior

The qodercli child inherits `QODER_AGENT_SDK_ENTRYPOINT`, interprets the invocation as Agent SDK mode, and rejects the ordinary CLI arguments before executing the evaluation prompt.

Removing the inherited variable allows the same evaluation to proceed through normal CLI startup:

```bash
env -u QODER_AGENT_SDK_ENTRYPOINT \
  skill-up run ./evals/eval.yaml
```

## Impact

The failure occurs when all of the following are true:

- `skill-up` is launched from Qoder Work or another Qoder Agent SDK-hosted process;
- the process inherits a non-empty `QODER_AGENT_SDK_ENTRYPOINT`;
- the evaluation uses the `qodercli` engine.

Ordinary terminals without this environment variable are unaffected.

The following qodercli versions were reproduced locally and all rejected the adapter invocation:

- 1.1.3
- 1.1.6
- 1.1.9

The earliest affected qodercli version has not been established. The earliest version verified in this reproduction is 1.1.3.

## Root cause

At commit [`acbbc74bb2fb029136f3994cb22b5d8720415738`](https://github.com/alibaba/skill-up/commit/acbbc74bb2fb029136f3994cb22b5d8720415738), the runtime builds the child environment from the complete host environment:

```go
envMap := envMapFromList(os.Environ())
maps.Copy(envMap, persistentEnv)
maps.Copy(envMap, callEnv)
```

Source:

https://github.com/alibaba/skill-up/blob/acbbc74bb2fb029136f3994cb22b5d8720415738/internal/runtime/runtime.go#L23-L37

The qodercli adapter then builds an ordinary headless CLI command:

```go
cmd := binary + " --permission-mode=bypass_permissions" + qoderJSONOutputFlag
cmd += " -p " + shellQuote(instruction)
```

Source:

https://github.com/alibaba/skill-up/blob/acbbc74bb2fb029136f3994cb22b5d8720415738/internal/agent/qodercli.go#L202-L229

`qoderRunEnvVars` does not create `QODER_AGENT_SDK_ENTRYPOINT`; the variable comes from the host process. However, the adapter also does not remove or override it before starting a new qodercli process.

This is an ambient-environment leakage issue that causes protocol mode confusion.

## Suggested fix

Sanitize Qoder Agent SDK-specific environment variables before starting qodercli from the ordinary CLI adapter. At minimum, the child process should not inherit `QODER_AGENT_SDK_ENTRYPOINT`.

The sanitization should cover every path that launches qodercli, including:

- initial `Run`;
- stdin/file prompt delivery;
- multi-turn `RunTurn`/resume;
- global and CN qodercli profiles where applicable.

Simply removing the variable from `qoderRunEnvVars` is not sufficient because it is inherited separately from `os.Environ()`.

Changing the command to only add:

```text
--input-format stream-json --output-format stream-json
```

would also be incomplete. Genuine Agent SDK mode requires structured stdin, initialization/control messages, and an SDK authentication payload. The existing adapter is an ordinary CLI adapter and should isolate itself from the parent SDK environment instead of accidentally entering SDK mode.

## Suggested regression test

Add an adapter or runtime integration test that:

1. sets `QODER_AGENT_SDK_ENTRYPOINT=sdk-ts` in the parent environment;
2. launches a fake qodercli binary through each qodercli execution path;
3. asserts that the child does not receive the SDK entrypoint variable;
4. verifies that the generated command remains the ordinary `-p --output-format json` invocation;
5. covers initial runs, file/stdin prompt delivery, resume, and the CN profile.

## Environment

- skill-up revision: `acbbc74bb2fb029136f3994cb22b5d8720415738` (`v0.10.0-14-gacbbc74`)
- qodercli versions reproduced: `1.1.3`, `1.1.6`, `1.1.9`
- qodercli 1.1.9 error: `sdk_invalid_args`
- Triggering host: Qoder Work / Qoder Agent SDK-hosted shell
- Control case: the error disappears when `QODER_AGENT_SDK_ENTRYPOINT` is unset

JHWang-1997:
## Summary

Prevent the built-in qodercli adapter from inheriting `QODER_AGENT_SDK_ENTRYPOINT` from SDK-hosted parent processes such as Qoder Work. Without this isolation, nested ordinary qodercli commands are misclassified as Agent SDK sessions and fail with `sdk_invalid_args`.

## Related issues

Closes #240

## Changes

- unset `QODER_AGENT_SDK_ENTRYPOINT` for version checks, inline and file/stdin runs, resumed sessions, and MCP remove/add commands
- apply the same isolation to the global and CN qodercli profiles
- add regression coverage for run, resume, profile-default, and MCP command builders
- document the behavior change in the changelog

## Test plan

- [x] `make test` passes
- [x] `make build` passes
- [x] Deterministic QoderCLI single-turn and multi-turn E2E tests with the SDK marker inherited
- [x] Manual run and MCP argument-validation checks with qodercli 1.1.3, 1.1.6, and 1.1.9

## Notes for reviewers

The adapter intentionally remains on qodercli's ordinary `-p --output-format json` protocol. Merely switching to `stream-json` would be incomplete because genuine Agent SDK mode also requires structured stdin, initialization/control messages, and an SDK authentication payload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,9 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Agent judges selecting a different provider no longer inherit runner credentials
   or endpoints, preventing judge keys from being sent to a runner endpoint.
-- QoderCLI runs now clear an inherited `QODER_AGENT_SDK_ENTRYPOINT` before
-  starting the CLI, preventing SDK-hosted parent processes such as Qoder Work
-  from forcing ordinary JSON invocations into the incompatible SDK protocol.
+- QoderCLI commands now clear an inherited `QODER_AGENT_SDK_ENTRYPOINT`,
+  preventing SDK-hosted parent processes such as Qoder Work from forcing
+  ordinary version checks, runs, resumes, and MCP configuration into the
+  incompatible SDK protocol.
 
 ## [0.10.0] - 2026-09-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Agent judges selecting a different provider no longer inherit runner credentials
   or endpoints, preventing judge keys from being sent to a runner endpoint.
+- QoderCLI runs now clear an inherited `QODER_AGENT_SDK_ENTRYPOINT` before
+  starting the CLI, preventing SDK-hosted parent processes such as Qoder Work
+  from forcing ordinary JSON invocations into the incompatible SDK protocol.
 
 ## [0.10.0] - 2026-09-01
 
--- a/internal/agent/mcp_test.go
+++ b/internal/agent/mcp_test.go
@@ -71,8 +71,8 @@ func TestBuildQoderMCPInstallCmd_Stdio(t *testing.T) {
 		t.Fatalf("buildQoderMCPInstallCmd failed: %v", err)
 	}
 	for _, want := range []string{
-		"qodercli mcp remove --scope project 'marker'",
-		"qodercli mcp add --scope project 'marker' -e 'MCP_TOKEN='\"$MCP_TOKEN\" -- 'node' '/tmp/marker-server.mjs'",
+		"env -u QODER_AGENT_SDK_ENTRYPOINT qodercli mcp remove --scope project 'marker'",
+		"env -u QODER_AGENT_SDK_ENTRYPOINT qodercli mcp add --scope project 'marker' -e 'MCP_TOKEN='\"$MCP_TOKEN\" -- 'node' '/tmp/marker-server.mjs'",
 	} {
 		if !strings.Contains(cmd, want) {
 			t.Fatalf("command missing %q:\n%s", want, cmd)
@@ -92,8 +92,8 @@ func TestBuildQoderMCPInstallCmd_CNStdio(t *testing.T) {
 		t.Fatalf("buildQoderMCPInstallCmdForBinary failed: %v", err)
 	}
 	for _, want := range []string{
-		"qodercn mcp remove --scope project 'marker'",
-		"qodercn mcp add --scope project 'marker'",
+		"env -u QODER_AGENT_SDK_ENTRYPOINT qodercn mcp remove --scope project 'marker'",
+		"env -u QODER_AGENT_SDK_ENTRYPOINT qodercn mcp add --scope project 'marker'",
 	} {
 		if !strings.Contains(cmd, want) {
 			t.Fatalf("CN command missing %q:\n%s", want, cmd)
--- a/internal/agent/qodercli.go
+++ b/internal/agent/qodercli.go
@@ -101,10 +101,10 @@ func NewQoderCLIAgent(cfg Config) *QoderCLIAgent {
 		cfg.CheckCmd = "command -v " + profile.binary
 	}
 	if cfg.VersionCmd == "" {
-		cfg.VersionCmd = profile.binary + " --version"
+		cfg.VersionCmd = qoderCLIInvocation(profile.binary) + " --version"
 	}
 	if cfg.RunCmd == "" {
-		cfg.RunCmd = profile.binary + " -p \"%s\" 2>&1"
+		cfg.RunCmd = qoderCLIInvocation(profile.binary) + " -p \"%s\" 2>&1"
 	}
 	if cfg.SkillPath == "" {
 		cfg.SkillPath = ".qoder/skills"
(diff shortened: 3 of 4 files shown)


JHWang-1997:
Validation record:

- `make fmt`: passed.
- `go test -race ./internal/agent -run 'Test(BuildQoder|QoderCLI)'`: passed.
- `make verify`: passed (`go vet`, `revive`, and `golangci-lint`; 0 issues).
- `make test`: passed across all packages with the host Qoder credential variables removed from the test environment:

  ```sh
  env -u QODER_PERSONAL_ACCESS_TOKEN \
    -u QODER_CN_ACCESS_TOKEN \
    -u QODERCN_PERSONAL_ACCESS_TOKEN \
    make test
  ```

- `make build`: passed.
- Manual argument-validation checks:
  - Before the fix, ordinary CLI invocations with `QODER_AGENT_SDK_ENTRYPOINT=sdk-ts` failed with `sdk_invalid_args` on qodercli 1.1.3, 1.1.6, and 1.1.9.
  - With the new `env -u QODER_AGENT_SDK_ENTRYPOINT` prefix, all three versions passed SDK-mode argument validation and reached the normal network/authentication stage; none produced `sdk_invalid_args`.

This confirms that the adapter remains on its existing ordinary JSON protocol while being isolated from SDK mode inherited from the parent process.

JHWang-1997:
Follow-up validation after commit `51b94db`:

- Completed the previously uncovered MCP path: both `qodercli mcp remove` and `qodercli mcp add` now clear `QODER_AGENT_SDK_ENTRYPOINT`.
- Applied the same isolation to the default version probe and fallback run command so every adapter-managed qodercli invocation uses the ordinary CLI protocol consistently.
- `make fmt`: passed.
- `go test -race ./internal/agent -run 'Test(NewQoder|BuildQoder|QoderCLI|BuildMCP)'`: passed.
- `make verify`: passed (`go vet`, `revive`, and `golangci-lint`; 0 issues).
- `make test`: passed across all packages.
- `make build`: passed.
- Deterministic QoderCLI single-turn and multi-turn E2E tests passed with `QODER_AGENT_SDK_ENTRYPOINT=sdk-ts` inherited by the skill-up process.
- Manual MCP argument-validation checks passed with qodercli 1.1.3, 1.1.6, and 1.1.9: after the wrapper removed the inherited marker, each version reached normal MCP command handling without `sdk_invalid_args`.

Risk review:

- No remaining adapter-managed qodercli execution path was found without SDK-environment isolation. The availability check uses `command -v`, and the installer invokes the official install script rather than qodercli itself.
- The wrapper remains compatible with the repository's supported POSIX-shell execution model, including Windows execution through Git Bash.
- Ordinary CLI authentication is unchanged: the command still requires the edition-specific PAT or existing local login state.
- The CN command builders are covered by regression tests; a qodercn binary was not available for a live local invocation.

zpzjzj (approved):
LGTM

PR #243, opened 2026-09-14
refactor(agent): retire legacy no-op config fields

zpzjzj:
## Summary

- keep `engine.entry` and `engine.model.params` loadable in v1alpha1 YAML, but drop them at the resolved-agent boundary with migration warnings
- remove the unused fields from resolved, capability, adapter, judge-inheritance, and reporting paths
- stop expanding environment references inside ignored model params
- document supported replacements and update stale proposal snippets

## Why

Both fields have existed since the initial public schema but have never been consumed by an adapter. Carrying them through runner, judge, and capability layers made requested configuration look like supported runtime behavior and would compound Judge Agent configuration ambiguity.

`engine.entry` should use `engine.custom.local.command` and `args`; adapter options should use `engine.kwargs` or `engine.custom.kwargs`.

## Compatibility

Existing v1alpha1 YAML still parses. A loader test covers the historical fields, while resolution tests cover warnings and ensure they do not reach adapters.

Part of #196. This PR does not introduce independent Judge Agent configuration or agent-native config ownership.

## Verification

- `go test -race ./internal/config ./internal/credential ./internal/agent ./internal/runner`
- `make verify`
- `make test`
- `make test-action`
- `make build`
- `make e2e`


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   unchanged unless the event log is explicitly enabled.
 
 ### Changed
+- Legacy no-op `engine.entry` and `engine.model.params` values are now dropped
```
JHWang-1997: [P2] Keep this change under Unreleased

v0.11.0 was already tagged and published before this commit, so placing this entry in its Changed section makes the release notes describe behavior that is not present in that release. Please move it under Unreleased, with a Changed subsection, so it is attributed to the next release.

zpzjzj: Good catch. I moved this entry into a new `Changed` subsection under `Unreleased` in 1215158, so the published v0.11.0 notes remain unchanged.

JHWang-1997 (approved):

PR #248, opened 2026-09-15
fix(action): extract pinned qodercli archive path

zpzjzj:
## Summary

- extract the pinned QoderCLI archive using its actual `./qodercli` member name
- install the extracted binary into `/usr/local/bin` with explicit executable permissions

## Root cause

Runner Image run 34938947910 downloaded the pinned QoderCLI 1.1.21 archive and passed SHA-256 verification, but requested the non-existent archive member `qodercli`. The archive contains `./qodercli`, so GNU tar exited with `Not found in archive`.

## Validation

- verified the pinned archive SHA-256
- extracted `./qodercli` and installed it with mode 0755
- `make verify`
- `make test`
- `make test-action`

Local full Docker image validation was unavailable because the Docker daemon is not running; the Runner Image workflow covers the Linux amd64 build path.


JHWang-1997 (approved):
LGTM

PR #250, opened 2026-09-16
feat(custom): support stateful multi-turn sessions

### LINKED ISSUE #249: feat(custom): support generic stateful multi-turn sessions
## Problem / motivation

Custom Engines can receive all configured `input.turns` messages in one `SessionInput`, but `CustomAgent` does not implement `SessionResumer`. The evaluator therefore falls back to one batch invocation and cannot provide true turn-by-turn behavior for custom integrations.

This prevents otherwise generic local or HTTP agent adapters from participating in:

- per-turn `post_condition` evaluation;
- `capture` and substitution into later turns;
- `skip_remaining` / early failure behavior;
- per-turn result reporting and judge assertions;
- native continuation of a server-side session without replaying the full conversation.

A live integration check with a stateful agent service confirmed that a wrapper can currently consume the batched `SessionInput.messages`, drive multiple requests against one remote session, and return a passing `SessionResult`. However, skill-up still emits the batch-fallback warning and cannot observe or gate the individual turns. This is a framework-level custom-engine limitation rather than an adapter-specific requirement.

## Proposed solution

Add an opt-in, transport-neutral stateful conversation contract for Custom Engines. The exact schema can be finalized during design, but it should allow `CustomAgent` to implement `SessionResumer` without adding any vendor-specific adapter.

One possible shape is a `custom.conversation_mode: stateful` option with these semantics:

1. Invoke the existing configured local command or HTTP endpoint once per user turn.
2. Include the current message and the prior `session_id` in the standard input contract.
3. Read the latest `session_id` from the standard `SessionResult` (the field already exists).
4. Feed that ID back into the next invocation, scoped to one case/configuration/attempt.
5. Return each invocation to the evaluator so existing `post_condition`, `capture`, early-stop, and per-turn reporting logic works unchanged.

The capability should be explicit. Existing Custom Engine configurations must keep their current batch behavior unless they opt in.

### Acceptance criteria

- Both `custom.transport: local` and `custom.transport: http` can opt into the same stateful multi-turn contract.
- The first turn receives no prior session ID; subsequent turns receive exactly the ID returned by the preceding `SessionResult`.
- A successful first turn with remaining turns but no `session_id` produces a clear execution error.
- Session state is isolated by case, configuration, retry attempt, and parallel worker; no "latest session" lookup is used.
- Existing timeout, environment/header rendering, kwargs, workspace, artifact, observability, and secret-masking behavior applies consistently on every turn.
- `post_condition`, `capture`, `skip_remaining`, and per-turn judge assertions work for opted-in Custom Engines.
- Existing custom configurations remain backward compatible and continue using batch fallback.
- Unit/E2E coverage includes local and HTTP transports, session-ID propagation, concurrent cases, missing IDs, non-zero/HTTP failures, and credential redaction.
- User-facing documentation and the engine support matrix describe both batch and stateful modes.

### Add a built-in adapter for each stateful agent

This does not scale for an open-source framework and couples skill-up to vendor-specific CLIs and APIs.

### Let each wrapper process all messages in one batch invocation

This can preserve remote conversational context internally, but skill-up cannot enforce inter-turn gates, capture dynamic values, stop early, or report individual turn outcomes.

### Replay full history through stateless HTTP requests

This works for some agents but does not cover services whose canonical continuation mechanism is an opaque server-side session ID, and it can be inefficient for long conversations.

## Additional context

Current `SessionResult` already exposes `session_id`, and the evaluator already owns the generic `SessionResumer` orchestration. The missing piece is a declarative way for `CustomAgent` to opt into and carry that protocol across its existing local/HTTP transports.

zpzjzj:
Custom Engines currently receive multi-turn cases as one batch, so services that maintain their own conversation session cannot evaluate capture, substitution, post-conditions, or per-turn assertions against a real resumed conversation. This adds opt-in `custom.conversation_mode: stateful`: skill-up invokes the configured local command or HTTP endpoint once per turn, passes the previous `SessionResult.session_id` through the next `SessionInput`, and keeps existing configurations on batch behavior.

The evaluator now treats stateful Custom Engine results as incremental, requires an intermediate session ID only when another turn will run, aggregates per-turn transcripts, usage, and artifacts, snapshots path artifacts before later turns can overwrite them, and reports the actual remaining timeout to each invocation. Configuration validation, design/user documentation, local and HTTP fixtures, concurrency coverage, and end-to-end tests are included.

Closes #249

Validation:

- `make verify`
- `make test`
- `go test -tags e2e -run 'TestPipeline_CustomEngine_Stateful(HTTP|Local)Transport' -v ./e2e`
- Full `make e2e`: both new stateful tests passed; three existing tests were terminated during parallel execution under local resource pressure, and those three passed together on an immediate targeted rerun.


internal/agent/custom.go
```diff
@@ -229,7 +252,7 @@ func (a *CustomAgent) prepareRun(rt Runtime, opts ExecOptions, messages []transc
 		return nil, fmt.Errorf("render custom.kwargs: %w", err)
 	}
 
-	sess := a.buildSessionInput(rt, opts, messages, renderedKwargs, timeoutSec)
+	sess := a.buildSessionInput(rt, opts, messages, sessionID, renderedKwargs, timeoutSec)
```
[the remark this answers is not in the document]
zpzjzj: Fixed in b2dae9b. Custom HTTP request bodies can now reference `${session_id}`: it is empty on the first turn and contains the prior `SessionResult.session_id` on subsequent turns. The variable is exposed only while rendering `custom.http.request_body`; command-line contexts continue to reject it. Added unit coverage for the explicit request body and session rotation.

internal/evaluator/evaluator.go
```diff
@@ -1357,7 +1357,10 @@ func sessionGeneratedFiles(sessionResult *agent.SessionResult) []string {
 	if sessionResult == nil || sessionResult.Artifacts == nil {
 		return nil
 	}
-	return sessionResult.Artifacts.GeneratedFiles
+	files := make([]string, 0, len(sessionResult.Artifacts.GeneratedFiles)+len(sessionResult.Artifacts.GeneratedFileSources))
+	files = append(files, sessionResult.Artifacts.GeneratedFiles...)
+	files = append(files, sessionResult.Artifacts.GeneratedFileSources...)
```
JHWang-1997, line 1369: Could we keep the original workspace paths out of the judge’s generated-files list? They’re needed for workspace-diff exclusion, but the same list is also passed to the judge. If two turns overwrite evidence.txt, the judge sees both turn snapshots plus the original path twice; those original entries now point to the second turn’s content. That could mislead turn-specific checks. The judge should receive only the snapshots, while the diff collector can still use the original paths.

zpzjzj: Fixed in 0745e1a and follow-up dae256e. The judge-facing generated-files list now contains only per-turn artifact snapshots; original workspace paths are retained in GeneratedFileSources and used only by workspace-diff exclusion. Framework-managed input/output JSON is also kept out of judge/report artifacts because it may contain a resumable session_id. Added coverage for the separate judge/diff views and framework-file exclusion. make verify, make test, and make e2e pass.

JHWang-1997 (approved):

PR #257, opened 2026-09-17
fix(agent): resolve symlinked skill sources and fail loudly on zero-file installs

### LINKED ISSUE #253: Skill source behind a symlink silently installs 0 files (filepath.Walk does not follow symlinked root)
## Summary

When a Skill source directory is a **symlink** (a very common layout — e.g. `~/.agents/skills/my-skill` pointing to a repo checkout elsewhere), `skill-up` silently installs **zero files** into the case workspace. The run proceeds, debug logs even print `Evaluator: skill installed: my-skill`, and the `with_skill` variant effectively runs *without* the skill — silently invalidating any with/without benchmark comparison.

## Root cause

`internal/agent/skill.go` → `ListSkillFiles` walks the source with `filepath.Walk(sourceDir, ...)`. Go's `filepath.Walk` does **not follow a symlinked root**: when `sourceDir` itself is a symlink, the walk visits only the link node. The selector records a single `"."` entry; `installSkill` then stats it, sees a directory, skips it — and uploads nothing, with no error.

Reproduction (against current `main`):

```go
files, _ := agent.ListSkillFiles("/path/to/symlinked-skill", nil, nil)
// => ["."]   (1 entry — the link node itself)

files2, _ := agent.ListSkillFiles("/path/to/symlinked-skill/", nil, nil)
// => [".", "SKILL.md", "references", "references/a.md", ...]  (correct)
```

Because `resolveSkillConfig` builds the source as `filepath.Join(skillDir, ref.Path)` (never a trailing slash), any skill reached through a symlink hits this path.

## Impact

- No error anywhere; the only symptom is an empty `<workspace>/<target>/` directory.
- Benchmark mode reports `with_skill` vs `without_skill` deltas that are actually baseline-vs-baseline. I burned several full eval runs before noticing the agent never saw the skill (verified via the engine's session logs: the skill name never appeared in context).

## Suggested fixes (either/both)

1. Resolve the source with `filepath.EvalSymlinks` before walking (or walk an explicitly resolved root).
2. Fail loudly when installation yields zero files — an install that uploads nothing is almost never intentional, and a warning/error would have surfaced this immediately.

## Environment

- skill-up `main` (commit 4135cc3), go1.25.6 darwin/arm64
- Custom engine (`engine.custom`, local transport), `environment.type: none`

BiBoyang:
## Summary

Fixes #253.

When a skill source directory is a symlink (e.g. `~/.agents/skills/my-skill` → a repo checkout elsewhere), `filepath.Walk` does not follow the symlinked root: the walk visits only the link node, the selector records just `"."`, and `installSkill` uploads **zero files with no error** — while debug logs still print `skill installed: <name>`. The `with_skill` variant then runs without the skill, silently invalidating benchmark comparisons.

## Changes

- `ListSkillFiles`: resolve the source root with `filepath.EvalSymlinks` before walking. Note the walk root and the `filepath.Rel` base both use the resolved path, so relative file keys are unaffected; `installSkill` still reads file contents through the original (possibly symlinked) source path, which the OS resolves per-file.
- `installSkill`: error out when the selected set uploads zero files. An install that uploads nothing is never intentional — the source is empty, unreadable, or the include/exclude filters matched nothing — and silently proceeding turns `with_skill` into a second baseline.

## Tests

New unit coverage in `internal/agent/agent_test.go`:

- `TestListSkillFiles_SymlinkedSourceDir` — a symlinked source lists `SKILL.md` and nested files
- `TestInstallSkill_SymlinkedSource` — end-to-end install through a symlink into a `NoneRuntime` workspace lands the file content
- `TestInstallSkill_ZeroFilesSelectedFails` — an empty source now errors instead of silently installing nothing

`go test -race ./...`, `go vet ./...`, and `gofmt` are all clean.

## Verification against the original report

The reporting setup (a skill suite where every skill directory is a symlink) was re-run against this fix: `skills/` in the case workspace now contains `SKILL.md` + `references/` for the `with_skill` variant and stays empty for `without_skill`, and the with/without benchmark delta became meaningful.


zpzjzj (approved):
LGTM

PR #260, opened 2026-09-18
fix(judge): harden agent_judge JSON output prompt

### LINKED ISSUE #259: fix(judge): strengthen first-attempt agent_judge prompt so agentic judges do not emit a prose prefix
## Summary

When `agent_judge` runs on an **agentic** judge engine (e.g. `qodercli`), the first judge call sometimes returns a natural-language message instead of the required JSON object. The parser then fails on the very first character, and `AgentJudge.Evaluate` burns a **full extra model call** on the built-in correction retry. This adds avoidable latency to every affected case.

We should tighten the **first-attempt** judge prompt so the model's *final* message is JSON-only, matching how strict the correction prompt already is. Scope of this issue is **prompt hardening only** (see "Out of scope" below).

## Evidence (from a real CI run)

Observed repeatedly in the `run-skill-up` step log (3 cases in one run: `vm-host-lookup`, `vm-health`, `nc-health`):

```
level=WARNING msg="agent_judge output failed validation; retrying once with correction guidance: agent_judge failed to parse agent output: invalid JSON response: invalid character 'I' looking for beginning of value"
level=DEBUG   msg="agent_judge correction retry produced a valid response"
```

`invalid character 'I' looking for beginning of value` means the judge's final message began with the letter `I` — i.e. prose such as "I have loaded the skill… / I'll now evaluate…", **not** `{`. This is a format/instruction-adherence failure, not truncation (truncation shows up as `invalid character '<'`). Every occurrence was recovered by the retry, so correctness was preserved, but each one cost a second full judge model call.

## Root cause

The judge prompt instructs an agentic judge to invoke Skill tools and read Review Materials before grading (`appendJudgeSkillInstructions`, `appendReviewMaterials`). After that tool-using turn, the model tends to close with a natural-language summary. That closing message becomes the captured `FinalMessage`, and it starts with prose.

The **first-attempt** prompt (`buildJudgePrompt`) only says:

```
IMPORTANT: Return only the required JSON object, optionally wrapped in one JSON code fence. Do not add prose or extra fields.
```
(`internal/judge/agent_judge.go` ~L785-786)

The **correction** prompt (`buildAgentJudgeCorrectionPrompt`) is much stricter and — crucially — does not re-ask for tool use, which is why the retry passes on the first try:

```
Do not add commentary, do not wrap the response in a Markdown fence, and do not introduce fields outside the schema.
Allowed root field: results.
Required result fields with exact casing: criterion_id, passed, evidence, failures.
```
(`internal/judge/agent_judge.go` ~L808-816)

So the asymmetry is the problem: the strictness that makes the retry succeed is missing from the first attempt.

## Proposed change (prompt only)

In `buildJudgePrompt` / `appendRequiredResponseFormat` (`internal/judge/agent_judge.go`):

1. Add an explicit **final-message** rule that survives tool use, e.g.:
   - "After you finish loading any judge Skills and reading the review materials, your **final message must contain ONLY the JSON object** — it must begin with `{` (or a single ```json fence) and end with `}` (or the closing fence)."
   - "Do **not** begin the final message with any prose, preamble, acknowledgement, or narration (e.g. 'I', 'Sure', 'Here is'). Do not add commentary before or after the JSON."
2. Port the correction prompt's field-casing line into the first-attempt prompt: "Required result fields with exact casing: `criterion_id`, `passed`, `evidence`, `failures`." Today that casing rule is stated **only** in `buildAgentJudgeCorrectionPrompt`.
3. Resolve the fence inconsistency: the first-attempt prompt allows "optionally wrapped in one JSON code fence" while the correction prompt says "do not wrap … in a Markdown fence". Pick one contract and state it identically in both.

Goal: make the first attempt as strict as the correction attempt so the correction retry becomes the exception rather than a routine second model call.

## Out of scope (tracked separately, not part of this fix)

- **Judge model override silently ignored on `qodercli`.** The run logged: `engine "qodercli" does not support model "qwen3.7-max"; the model override is omitted and local Qoder settings will be used`. The configured judge model never takes effect, which likely worsens format adherence. Consider surfacing this at config-validation time instead of a runtime warning.
- **Genuine judge timeouts.** A separate case (`vm-start-stop-diagnose`) logged `qodercli run failed: context deadline exceeded (judge.timeout_seconds=0, parent_ctx_expired=true)` — a different failure mode handled via case/judge timeout config, not the prompt.

## Related

- #192 (strict program-owned output contract for agent_judge) — this issue is the prompt-side follow-up to reduce how often that contract is violated on the first attempt.

JHWang-1997:
## Summary

- Share a strict final-message JSON contract between first-attempt and correction judge prompts.
- Require the final judge message to contain only the JSON response after any tool use, preventing prose prefixes such as `I ...` from triggering a full correction retry.
- Add regression coverage that keeps the two prompt contracts aligned.

Fixes #259

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/judge/`


zpzjzj (approved):
LGTM

PR #262, opened 2026-09-19
fix(qodercli): pass through explicit model names and IDs

### LINKED ISSUE #247: feat(qodercli): pass through explicit model names and IDs
## Problem / motivation

Qoder CLI supports selecting a concrete model through the --model option. In addition to the built-in tiers, current versions can accept models returned by --list-models and custom model IDs.

skill-up still limits Qoder CLI models to five hard-coded values:

- lite
- efficient
- auto
- performance
- ultimate

Any other value is rejected during adapter configuration and removed again by QoderCLIAgent.appliedModelName. The generated command therefore omits --model and falls back to Qoder CLI's local default.

For example:

- engine.name: qodercli
- engine.model.provider: qoder
- engine.model.name: qwen3.7-plus

The expected command includes --model qwen3.7-plus. The current command omits it. This prevents evaluations from selecting newly available models and may cause them to run with a different model than requested.

## Proposed solution

- Treat any non-empty Qoder CLI model value as an opaque model name or ID and pass it through to --model after trimming whitespace.
- Apply the same behavior to initial runs, stdin-based runs, and resumed sessions.
- Keep an omitted model as "use the Qoder CLI default."
- Preserve the existing tier values without behavior changes.
- Let Qoder CLI validate whether the selected model is available to the current account instead of maintaining a static allowlist in skill-up.
- If Qoder CLI rejects an unavailable or unauthorized model, surface the error clearly and include the requested model.
- Never omit an explicitly requested model or retry with the local default.
- Keep the existing provider, authentication, base URL, and edition behavior unchanged.

## Model availability validation

Model availability is account-specific and server-driven, so skill-up should not replace the current tier allowlist with another static list.

For the initial implementation, Qoder CLI should remain the source of truth. skill-up should forward the requested model unchanged and preserve any error returned by Qoder CLI.

A future optional preflight may query qodercli --list-models once per evaluation if Qoder provides stable machine-readable output. A failed model-list query must not be treated as proof that the model is unavailable because it can also fail due to authentication or network problems.

## Compatibility

Existing configurations remain unchanged when the model is omitted or set to lite, efficient, auto, performance, or ultimate.

The only intentional behavior change is for an explicit value that skill-up currently ignores. After this change, that value is passed to Qoder CLI and either takes effect or fails clearly instead of silently falling back to the default model.

## Alternatives considered

Extending the static allowlist whenever Qoder adds a model would quickly become outdated because the available model list is server-driven and custom model IDs cannot be enumerated in skill-up.

Continuing to omit unsupported values is unsafe for evaluations because the run may silently use a different model.

## Acceptance criteria

- Models configured through eval YAML or skill-up run --model are forwarded to Qoder CLI.
- Initial, stdin-based, and resumed sessions use the requested model.
- Existing tier configurations and omitted-model configurations continue to work unchanged.
- An unavailable or unauthorized model produces a clear error containing the requested model.
- skill-up never falls back to Qoder CLI's default after an explicit model was requested.
- Requested and applied model fields report the forwarded value.
- Unit tests cover a tier value, a concrete model name, a custom model ID, and resumed sessions.
- English and Chinese Qoder CLI documentation no longer state that only the five tiers are supported.
- If model-ID passthrough requires a minimum Qoder CLI version, document that requirement and surface an actionable compatibility error.

## Additional context

The current restriction is implemented in supportedQoderModels, ModelPolicyQoderTier, and QoderCLIAgent.appliedModelName.

Related configuration work: https://github.com/alibaba/skill-up/issues/196

Qoder CLI model documentation: https://docs.qoder.com/cli/model

Qoder CLI option reference: https://docs.qoder.com/cli/cli-reference

JHWang-1997:
## Summary
- Remove both Qoder model allowlists and forward trimmed, case-preserving model names/IDs across initial, stdin, and resumed calls.
- Preserve historical tier passthrough and omitted-model defaults. Keep Global/CN authentication and provider/API key/base URL isolation unchanged.
- Include explicit model context in execution errors without falling back to a default model; preserve requested/applied reporting.
- Update English/Chinese documentation, the changelog, and skill scaffolding expectations.

Closes #247

## Verification
- `make fmt`
- `make verify` (0 issues)
- `make test` (`go test -race ./...`)
- Review rerun: `go test -race ./internal/agent ./internal/credential ./internal/evaluator ./internal/judge`
- `git diff --check`
- Commit hooks: go vet, revive, incremental golangci-lint (0 issues)

Regression tests cover historical tiers, concrete and slash-containing model IDs, whitespace, shell quoting, Global/CN command generation, inline/stdin/resume paths, invalid-model/auth/runtime errors, requested/applied metadata, judge resolution, and evaluator retries retaining the explicit model.

## Compatibility and Remaining Validation
- skill-up forwards historical tiers, including lite; actual availability is determined by the installed Qoder CLI and account.
- Slash-containing CLI model IDs can use an explicit provider to avoid legacy provider/model parsing; YAML engine.model.name preserves the full ID.
- Live CN, valid custom/BYOK models, and older Qoder versions have not been verified end to end. Mock coverage does not claim upstream availability.
- No minimum Qoder version is asserted without verification.


zpzjzj (approved):
LGTM

